PR #17, opened 2020-03-25
Remove non staff meqs module

### LINKED ISSUE #16: Remove any MEQS tag not marked with security level staff

NeunEinser:
Resolves #16

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,6 +9,7 @@ version = "1.0-SNAPSHOT"
 
 repositories {
     mavenCentral()
+    maven("https://maven.ihtsdotools.org/content/repositories/releases/")
 }
 
 val logBackVersion = "1.2.3"
@@ -26,7 +27,7 @@ dependencies {
     }
 
     implementation("com.uchuhimo", "konf", "0.22.1")
-    implementation("net.rcarz", "jira-client", "0.5")
+    implementation("net.rcarz", "jira-client", "0.6.3-IHTSDO")
     implementation("org.slf4j", "slf4j-api", "1.7.25")
     implementation("ch.qos.logback", "logback-classic", logBackVersion)
     implementation("ch.qos.logback", "logback-core", logBackVersion)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,7 +9,7 @@ version = "1.0-SNAPSHOT"
 
 repositories {
     mavenCentral()
-    maven("https://maven.ihtsdotools.org/content/repositories/releases/")
+    maven("https://jitpack.io")
 }
 
 val logBackVersion = "1.2.3"
@@ -27,7 +27,7 @@ dependencies {
     }
 
     implementation("com.uchuhimo", "konf", "0.22.1")
-    implementation("net.rcarz", "jira-client", "0.6.3-IHTSDO")
+    implementation("com.github.rcarz", "jira-client", "master-SNAPSHOT")
     implementation("org.slf4j", "slf4j-api", "1.7.25")
     implementation("ch.qos.logback", "logback-classic", logBackVersion)
     implementation("ch.qos.logback", "logback-core", logBackVersion)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,7 +14,7 @@ repositories {
 
 val logBackVersion = "1.2.3"
 val arrowVersion = "0.10.4"
-val kotlinTestVersion = "3.4.2"
+val kotestVersion = "4.0.1"
 
 dependencies {
     implementation(kotlin("stdlib-jdk8") as String) {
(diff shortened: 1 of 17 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
```diff
@@ -46,6 +46,17 @@ fun addComment(issue: Issue, comment: String) = runBlocking {
...
+// The used library doesn't include visibility information in comments for whatever reason
+fun isCommentRestrictedTo(jiraClient: JiraClient, comment: Comment, group: String) = runBlocking {
+    Either.catch {
+        val commentJson = jiraClient.restClient.get(URI(comment.self)) as JSONObject
+        if (commentJson.contains("visibility")) {
+            val visibility = commentJson.getJSONObject("visibility")
+            visibility.get("type") == "group" && visibility.get("value") == group
+        } else false
+    }
+}
```
NeunEinser: I really don't like the jira library used here.
I already didn't like having to take care of the json for updating comments, and now it turns out the library classes don't actually include all the information returned by the API.

NeunEinser:
Should this module also post a message when taking action?

urielsalis:
Maybe leave a staff message with the name of the user who made the comment?

src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
urielsalis: https://github.com/rcarz/jira-client/blob/master/src/main/java/net/rcarz/jiraclient/Comment.java#L41

It does allow to get visibility from the comment, you might need to get the comment by I'd again or add it to the expand so jira returns the info

You can also do .update() to change the visibility

src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
```diff
@@ -61,6 +72,22 @@ fun updateCommentBody(jiraClient: JiraClient, comment: Comment, newValue: String
     }
 }
 
+fun restrictComment(jiraClient: JiraClient, comment: Comment, group: String, body: String? = comment.body) = runBlocking {
+    Either.catch {
```
urielsalis: You can do theComment.update() and change the visibility

NeunEinser:
> Maybe leave a staff message with the name of the user who made the comment?
Not sure if that's necessary.
The original comment will already be left with staff visibility, but instead of MEQS_WHATEVER, the tag will be replaced by _WHATEVER excluding the MEQS prefix.

src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
NeunEinser: The link you posted seems to be a newer version of the library. I already looked at the comment class, it does not contain that method.

src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
NeunEinser: This method does not exist in the version being used.

urielsalis: We should be able to update it by just changing the version in build.gradle

NeunEinser: The latest version available at maven seems to be https://mvnrepository.com/artifact/net.rcarz/jira-client/0.6.3-IHTSDO.
This version does include `Comment.update()` but still no visibility.

urielsalis: ugh, I checked the repo, apparently they never published a newer version. The other libs I could find was the official one that relies on everything being a rest client call, so I would just leave it like this until we can find another replacement library

urielsalis: Or build the lib from sourve

NeunEinser: > Or build the lib from sourve
Not sure

NeunEinser: > Or build the lib from sourve

Not sure how this would work with gradle.

urielsalis: Done using jitpack

build.gradle.kts
```diff
@@ -27,7 +27,7 @@ dependencies {
     }
 
     implementation("com.uchuhimo", "konf", "0.22.1")
-    implementation("net.rcarz", "jira-client", "0.6.3-IHTSDO")
+    implementation("com.github.rcarz", "jira-client", "master-SNAPSHOT")
```
Marcono1234: @urielsalis, shouldn't this use a specific version (or commit hash)? Otherwise that might cause unexpected build failures in the future.

urielsalis:
The lib hasnt been touched since 2018, so I dont think it will be a issue.
But yes, will change in a future pr

PR #18, opened 2020-03-26
Empty module

### LINKED ISSUE #10: Port empty module
https://github.com/mojira/arisa-js/blob/master/checks/ModuleEmpty.js

urielsalis:
Fixes #10

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
@@ -46,18 +46,10 @@ fun addComment(issue: Issue, comment: String) = runBlocking {
     }
 }
 
-fun resolveAsInvalid(issue: Issue) = runBlocking {
+fun resolveAs(issue: Issue, resolution: String) = runBlocking {
     Either.catch {
         issue.transition()
-            .field(Field.RESOLUTION, "Invalid")
-            .execute("Resolve Issue")
-    }
-}
-
-fun resolveAsIncomplete(issue: Issue) = runBlocking {
-    Either.catch {
-        issue.transition()
-            .field(Field.RESOLUTION, "Incomplete")
+            .field(Field.RESOLUTION, resolution)
             .execute("Resolve Issue")
     }
 }
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
@@ -54,6 +54,14 @@ fun resolveAsInvalid(issue: Issue) = runBlocking {
     }
 }
 
+fun resolveAsIncomplete(issue: Issue) = runBlocking {
+    Either.catch {
+        issue.transition()
+            .field(Field.RESOLUTION, "Incomplete")
+            .execute("Resolve Issue")
+    }
+}
+
 fun updateCommentBody(comment: Comment, body: String) = runBlocking {
     Either.catch {
         comment.update(body)
--- a/src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
@@ -1,0 +1,53 @@
+package io.github.mojira.arisa.modules
+
+import arrow.core.Either
+import arrow.core.extensions.fx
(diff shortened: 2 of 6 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
```diff
@@ -54,6 +54,14 @@ fun resolveAsInvalid(issue: Issue) = runBlocking {
     }
 }
 
+fun resolveAsIncomplete(issue: Issue) = runBlocking {
+    Either.catch {
+        issue.transition()
+            .field(Field.RESOLUTION, "Incomplete")
+            .execute("Resolve Issue")
+    }
+}
```
NeunEinser: Maybe it would make sense to merge resolveAs methods to one with a string argument and pre-fill that argument using `partially`?

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
```diff
@@ -0,0 +1,53 @@
...
+
+data class EmptyModuleRequest(
+    val numAttachments: Int,
+    val description: String?,
+    val environment: String?
+)
+
+const val DESCDEFAULT = "Put the summary of the bug you're having here\r\n\r\n*What I expected to happen was...:*\r\nDescribe what you thought should happen here\r\n\r\n*What actually happened was...:*\r\nDescribe what happened here\r\n\r\n*Steps to Reproduce:*\r\n1. Put a step by step guide on how to trigger the bug here\r\n2. ...\r\n3. ..."
+const val ENVDEFAULT = "Put your operating system (Windows 7, Windows XP, OSX) and Java version if you know it here"
+const val MINLENGTH = 5
```
NeunEinser: Wouldn't those make sense as config keys?

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
```diff
@@ -0,0 +1,53 @@
...
+
+class EmptyModule(
+    val resolveAsIncomplete: () -> Either<Throwable, Unit>,
+    val addEmptyComment: () -> Either<Throwable, Unit>
+) : Module<EmptyModuleRequest> {
+    override fun invoke(request: EmptyModuleRequest): Either<ModuleError, ModuleResponse> = Either.fx {
+        with(request) {
```
NeunEinser: Oh `with` is some neat syntax sugar I wasn't aware of yet! Nice!

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
urielsalis: I was thinking about it, but at the end we dont even have this defaults anymore, so I dont think this will ever change

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
urielsalis: Kotlin is full of this syntax sugar :D

src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
urielsalis: Good idea

urielsalis: Done

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
urielsalis: Note it only goes as far as Windows 7

NeunEinser: Well, Mojang could update it at any point.
Is this even the same across all projects? It would be kinda weird for the BE.

For me, the default text is empty anyways, I'm not quite sure why.

urielsalis: I think they removed the default text,  this is a copy from the old module but im probably going to remove it as soon as I verify that (post launch)

NeunEinser: Ah yes, that does make sense then

PR #20, opened 2020-03-26
Change to json config to allow for proper lists

NeunEinser:
This converts the external config file to json.
This allows for using proper lists.
The config should now look similar to this:

```json
{
	"arisa": {
		"credentials": {
			"username": "arisa",
			"password": "1234"
		},
		"issues": {
			"projects": [
				"MCTEST"
			]
		},
		"modules": {
			"attachment": {
				"extensionBlacklist": [
					"test"
				]
			}
		}
	}
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@
 /build/
 .idea/
 arisa.yml
+arisa.json
 /.kotest
 /out/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,5 +3,6 @@
 /build/
 .idea/
 arisa.yml
```
Marcono1234: Is this `.gitignore` entry still needed then?

NeunEinser: I kept it intentionally, so that no one accidentally commits credentials from an old *.yml.

PR #21, opened 2020-03-27
Add CrashModule

### LINKED ISSUE #9: Port crash module
https://github.com/mojira/arisa-js/blob/master/checks/ModuleCrash.js

NeunEinser:
Resolves #9

This ports the crash module from arisa-js, so far without any change in logic. There are a couple of things that seemed a little bit questionable to me though.

https://github.com/mojira/arisa-kt/blob/8dacad245e5237ced0039d3549f75975396f453d/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt#L141-L148
The crash module collects some information from different sources into a CrashInfo object.
However, while they are written, there is nothing reading or relying on `minecraftVersion` or `javaVersion`.

https://github.com/mojira/arisa-kt/blob/8dacad245e5237ced0039d3549f75975396f453d/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt#L96
This is the one that's a bit harder to solve.
In here, it is just blatantly assumed that the 7th line (index 6) of every crash report contains the exception.
Even ignoring the fact Mojang may change the layout of a crash report at any time, since not just attached crash reports, but also attached launcher log files or a crash report added inside of the issue's description are evaluated, this already breaks there.
This means we probably should think of criteria of how the line with the exception header should look like and than search for that line similarily as how other bits of information are extracted.

Finally, running `gradlew build` locally and in the github check is broken for me on this branch.
For some reason, gradle is unable to resolve an import for `io.kotest.matchers.booleans.shouldBeTrue` in `CrashModuleTest`, while IntelliJ can.


src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
```diff
@@ -0,0 +1,600 @@
...
+            listOf(CrashDupeConfig("minecraft", "Pixel format not accelerated", "MC-297")),
+            10
+        )
+        val modded = mockAttachment("crash_modded.txt", Calendar.getInstance().time, EXAMPLE_CRASH.toByteArray())
+        val dupe = mockAttachment("crash_dupe.txt", Calendar.getInstance().time, EXAMPLE_CRASH.toByteArray())
+        val request = CrashModuleRequest(listOf(modded, dupe), "", Calendar.getInstance().time)
+
+        val result = module(request)
+
+        result.shouldBeRight(ModuleResponse)
```
urielsalis: Minor, you can do result shouldBeRgiht ModuleResponse (spaces instead of . and () )

NeunEinser: I just copied it from other classes, that's how it's done everywhere else in the project :D

NeunEinser:
@urielsalis Do you think the logic I ported from the old module which seemed a little odd are okay for this project?

NeunEinser:
@urielsalis Do you think the logic I ported from the old module which seemed a little odd, which I mentioned in my initial comment, are okay for this project?

urielsalis:
To start, I would like it to be 1-1. We can start refactoring now

PR #27, opened 2020-04-03
fix CHK processing unconfirmed tickets

NeunEinser:


src/main/kotlin/io/github/mojira/arisa/infrastructure/Jira.kt
```diff
@@ -12,14 +12,23 @@ import net.rcarz.jiraclient.Version
...
     Either.catch {
         issue
             .transition()
-            .field(chkField, Instant.now().toString().replace("Z", "-0000"))
+            .field(
+                chkField,
+                Instant
+                    .now()
+                    .with(ChronoField.NANO_OF_SECOND, 0)
+                    .with(ChronoField.MILLI_OF_SECOND, 123)
```
Marcono1234, line 28: Why change this value to 123?

urielsalis: Was commented in the channel, probably should have a comment here. Jira API doesnt support the full ISO format, only accepts it when it has 3 decimals after seconds. On some systems you always get 3, on others 4, on others 6. This makes sure you always only use 3 (and replace the Z from ISO to -0000 for same reason)

Marcono1234: Ah thanks, I have read the part about offset `-0000` instead of `Z`, but must have missed the other part.

PR #39, opened 2020-04-06
add confirmation module

### LINKED ISSUE #22: Set tickets back to unconfirmed if it wasnt done by Mojang, Mod or Helper

NeunEinser:
Resolves #22

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RevokeConfirmationModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RevokeConfirmationModule.kt
@@ -1,0 +1,45 @@
+package io.github.mojira.arisa.modules
+
+import arrow.core.Either
+import arrow.core.extensions.fx
+import net.rcarz.jiraclient.ChangeLogEntry
+
+data class RevokeConfirmationModuleRequest(
+    val confirmationStatus: String,
+    val changeLog: List<ChangeLogEntry>
+)
+
+class RevokeConfirmationModule(
+    private val getGroups: (String) -> Either<Throwable, List<String>>,
+    private val setConfirmationStatus: (String) -> Either<Throwable, Unit>,
+    private val confirmationFiledId: String
+) : Module<RevokeConfirmationModuleRequest> {
+    override fun invoke(request: RevokeConfirmationModuleRequest): Either<ModuleError, ModuleResponse> = with(request) {
+        Either.fx {
+            val confirmationChanges = changeLog.filter(::isConfirmationChange)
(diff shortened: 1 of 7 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/RevokeConfirmationModule.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    private fun getConfirmation(entry: ChangeLogEntry) =
+        entry.items.lastOrNull { it.field == confirmationFiledId }?.toString
+
+    private fun changedByVolunteer(entry: ChangeLogEntry): Boolean {
+        val groups = getGroups(entry.author.name)
+
+        return if (groups.isLeft())
+        // when in doubt, assume change was done by a volunteer
+            true
```
NeunEinser, line 41: Not entirely sure what should be done here.
It's kinda annoying to have a filter method that can fail, and a case that we didn't have to deal with yet. It would maybe a bit better to somehow collect these errors, but the filter method can't easily deal with Eithers.

urielsalis: You can do maps and filters inside eithers (they are monads) to combine them if needed

src/main/kotlin/io/github/mojira/arisa/modules/RevokeConfirmationModule.kt
```diff
@@ -0,0 +1,45 @@
...
+        else
+            (groups as Either.Right).b.any { it == "helper" || it == "global-moderators" || it == "staff" }
```
urielsalis, line 43: Not all Mojang staff has the helper, global-moderators or staff role :/

NeunEinser: Can you give me an example?
To me knowledge, everyone has at least staff.

urielsalis: everyone has staff, not all have helper. Perfect

src/main/kotlin/io/github/mojira/arisa/modules/RevokeConfirmationModule.kt
NeunEinser: I'm not quite sure how you mean that.
I don't see how combining a map with a filter would help here, or what Either has to do with that.

For this, it would be cool to have a filter in combination with an Either, so that the result is left when the filter function fails at some point for an entry of the list, but I don't see how to do that.
There might be other functional constructs I'm not aware of that are made for this case, of course.

PR #42, opened 2020-04-06
Fix piracy signatures using whitespaces not working

NeunEinser:
See https://github.com/mojira/arisa-kt/commit/9cce384da44760e3aed55b31183f37aea04546dc#r38319248

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/PiracyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/PiracyModule.kt
@@ -27,7 +27,7 @@ class PiracyModule(
     }
 
     private fun assertContainsSignatures(piracySignatures: List<String>, matcher: String) = when {
-        piracySignatures.any { matcher.split("\\s+".toRegex()).contains(it) } -> Unit.right()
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/PiracyModule.kt
```diff
@@ -27,7 +27,7 @@ class PiracyModule(
     }
 
     private fun assertContainsSignatures(piracySignatures: List<String>, matcher: String) = when {
-        piracySignatures.any { matcher.split("\\s+".toRegex()).contains(it) } -> Unit.right()
+        piracySignatures.any { """(?:\s|^)${Regex.escape(it)}(?:\s|$)""".toRegex().containsMatchIn(matcher) } -> Unit.right()
```
Pokechu22, line 30: `\b` for word boundary probably would have been better instead (though I think it's equivalent).

NeunEinser: Ah nice, changed now.

PR #44, opened 2020-04-06
add cache

### LINKED ISSUE #25: Cache tickets that already were checked for 5 minutes
Cache for 5 minutes tickets that were already checked to prevent rechecking them all the time.

Only for 5 minutes so if there is any other update it will still be picked up in the next run

NeunEinser:
Resolves #25


NeunEinser:
What do you mean with "reset each time" / what does "each time" refer to?

This will only happen when the entry is not in the cache already, because this happens after the filter, so a key in the cache won't be re-scheduled.
https://github.com/mojira/arisa-kt/blob/0c9ec1659098d4623118b345865603409c155f05/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt#L101-L105

PR #46, opened 2020-04-07
Added keep private module

### LINKED ISSUE #19: Mark tickets with MEQS for private as private again
If a ticket is make public but it still has MEQS for private, mark it as private again

NeunEinser:
Resolves #19


NeunEinser:
If you are talking about the config:
https://github.com/mojira/arisa-kt/blob/6240a51ad59d657e67af02da56445a4edb3492c9/src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt#L26-L32

I don't know why, but MCL and MCAPI are the only projects that have their own security level (`Minecraft Launcher - Private` and `Minecraft API - Private` respectively). All other projects use the same private security level with id `10318` simply called `Minecraft - Private`. I checked all of them and didn't want to list all projects if most of them use the same security level anyways.

In case this ever changes, we can just change the config.

Other then this I don't see where to add other projects or where other projects might not be supported.

PR #67, opened 2020-04-09
Module specific JQL

### LINKED ISSUE #59: Enable a custom search query per module
As mentioned in #38 it might be benefitial to enable the use of a custom search query per module, instead of all modules going over tickets from the same query,

This can have following advantages:
* In case a custom search query can be used, a new module can also go over old tickets, ignoring when the module was updated
* In case of a downtime, all modules that can use a custom query could catch back up.
* New modules might require to go over tickets that are currently not considered, either because they are old, or because they have a resolution or other restriction that is currently not considered for all modules.
* It could reduce overall processing time as certain modules just check tickets that matter to them

NeunEinser:
This is a refactor of how ArisaMain works and moves most of the logic to a new Module Executor.
This also removes most config values from the kotlin file and moves them to the json.
The json is now included in the commit, and arisa.yml is read again for credentials.

### Moving to JSON
There are several reasons I decided to do this in this PR.
- The way default values work, is that you can only default per value.
  - This means only people with access to the server files can change inherited values. Before this PR that was module-specific whitelists
  - This PR adds a list of resolutions and a jql query to this list as well now.
    - If this was not changed, these values could only be changed by the maintainer, and by no one else, which would suck.
- I wasn't very fond of defining every value as default value from the start.
- Especially as proven by Greymagic's recent PR it's harder to edit the config for unexperienced people in a kt file, as there is much clutter that's not really needed for configuring stuff.
- With this config it was pretty much encouraged to change default values rather than values in an actual config.
  - This means a rebuild of the entire project when changing the config, rather than just a restart.
- **To make this effective, arisa.json is included in the commit now**
  - **You MUST move credentials to `arisa.yml` now.**

### ModuleExecutor
The module executor is the central unit responsible for executing all modules now.
The logic in the module executor is mostly a mirror of what was in ArisaMain beforehand.

Changes include:
- All modules are now initialized only once. before, they were initialized for every issue.
- Most modules require a function that takes an issue and does something with it.
  - Since Modules are now initialized only once, this means the issue can not be substituted with partially anymore.
  - Because of this, most modules now require an issue to be sent in the request, which is then applied to the function in the module.
  - Some values that are sent in a request are technically available in the issue, however, I decided to **not** remove those.
    - While some values are available with simple dot syntax (e.g. `issue.comments`), others require something like `it.getField("environment") as? String?`, or in extreme cases even `((it.getField(config[Arisa.CustomFields.confirmationField])) as? JSONObject)?.get("value") as? String?`. I did not want modules to have to deal with this stuff.
- Every module is now executed separately in order to implement the option to have a per-module-jql.
  - To reduce network traffic, the result of every jql request is cached for every run, so if multiple modules use the same jql, only one request is made.
  - This also means execution order changed, which has consequences for the order in which stuff is logged as well.
    - The log would previously group issues and list all module responses per issue. Now it's instead grouped by module and all responses from all tickets a module was executed on

Some of the changes I implemented later are currently untested and with jira being down for an update, I can't test them now. I wanted to get this PR out now though, so people can take a look at it and tell me whether they like the direction I went with this.
Feel free to merge new modules or other changes first into master. I will take care of merging them into this PR.

Resolves #59


urielsalis:
> To make this effective, arisa.json is included in the commit now
> You MUST move credentials to arisa.yml now.

Does that mean I should move the credentials to arisa.yml and replace the arisa.json with the one in the repo?

urielsalis:
> All modules are now initialized only once. before, they were initialized for every issue.
The main reason for that was to pass the issue to the jira methods, that way modules didnt need to know about jira itself to operate, only about the data that we extracted out of it in ArisaMain. 
It made testing the module way easier as we didnt need to mock anything. It hasnt been really followed with attachment and future modules but I though it was a nice thing to have and that we can look into maybe?

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,304 @@
...
+    private val ticketCache = mutableListOf<String>()
+    private val ticketTimer = Timer("RemoveCachedTicket", true)
+
+    fun execute() {
+        // Cache issues returned by a query to avoid searching the same query for different modules
+        val queryCache = mutableMapOf<String, List<Issue>>()
+        val processedTickets = mutableMapOf<String, Boolean>()
+        val exec = ::executeModule.partially2(queryCache).partially2(processedTickets)
+
+        exec(Arisa.Modules.Attachment) {
```
urielsalis, line 79: It would be really nice if we can start adding some tests using the fact that we separated this into a separate file.

That means testing the filter logic we have in execute module, testing the ticket caching, that X module was executed if passed

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,304 @@
...
+                    (issue.comments.isNotEmpty() && issue.comments.last().updatedDate > latestChange.created) // And there is no comment posted after that
+            }
+        queryCache[combinedJql] = issues
+
+        issues
+            .filter { config[moduleConfig.whitelist] == null || config[moduleConfig.whitelist]!!.contains(it.key) }
+            .filter { config[moduleConfig.resolutions].contains(it.resolution.name) }
+            .map { it.key to executeModule(it) }
+            .forEach { (issue, response) ->
+                when (response.second) {
```
urielsalis, line 219: Can be response.second.fold(
  {//left part},
  {//right part}
)

That also automatically casts stuff so you dont need the with

NeunEinser:
> > To make this effective, arisa.json is included in the commit now
> > You MUST move credentials to arisa.yml now.
> 
> Does that mean I should move the credentials to arisa.yml and replace the arisa.json with the one in the repo?

When merging, yes. master currently does not read `arisa.yml`, but you could prepare already by just having both files so the merge is gonna be more smooth,
arisa.yml is in the current gitignore.

NeunEinser:
> > All modules are now initialized only once. before, they were initialized for every issue.
> 
> The main reason for that was to pass the issue to the jira methods, that way modules didnt need to know about jira itself to operate, only about the data that we extracted out of it in ArisaMain.
> It made testing the module way easier as we didnt need to mock anything. It hasnt been really followed with attachment and future modules but I though it was a nice thing to have and that we can look into maybe?

I agree that it would be cool to not have to work on API classes directly, I'm just not really sure how it could be done effectively.
We could ofc make our own Classes for Issue, comment and such and provide some access to them, but I'm not really sure about that. Do you have an idea for that as well?

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
NeunEinser: Gonna look into that later

urielsalis:
> We could ofc make our own Classes for Issue, comment and such and provide some access to them, but I'm not really sure about that. Do you have an idea for that as well?

That was my main approach when doing it, on top of passing issue and attachment to the methods so the jira api can operate directly on top of them without our module code having to know about it

urielsalis:
The main issue with that in the old code was the delete attachemnt module. I would probably refactor that one to receive a map of attachemnt name->funciton to delete itself. And that function is created from the jira function and passed without needing the actual attachemnt

violine1101:
It might make sense to consider using [TOML](https://github.com/toml-lang/toml) or YAML instead of JSON for the config file, as both of these are more human-readable, especially for multi-line strings (like we have them a lot), and don't require quotation marks all over the place.

NeunEinser:
The last push is not gonna build.
I changed the modules to not use JIRA api classes now, but did not change anything actually using the modules yet, so stuff using them is broken, this includes test and ModuleExecutor.

I'm going to bed now though. Feel free to tag a look at how I approached it in the modules 😉

NeunEinser:
Modules are now no longer exposed to library classes.

Most of them have their own data classes for needed objects now, and library classes are mapped in ModuleExecutor.
These classes include any needed data by the module, but not more than that.
They also include functions that for API calls, such as removing an attachment or adding an affected version.

In order to avoid class name clashes, those are declared within the Module class, so that when multiple classes define the same data classes, they don't clash.
For consistency, Request classes are now moved into the Module as well and are renamed simply to Request now (`AttachmentModuleRequest` -> `AttachmentModule.Request`, pay attention to the dot.)

I still haven't looked into moving to yaml or taml. Yaml is what we originally had, but there might have been problems when using lists, although looking at the spec it does seem like yaml supports lists and maps/dictionaries, so not sure about that one.
json is what was there already, just not committed.
I wouldn't mind switching to something which supports stuff like multiline texts, though.

I also didn't look into tests further for now.

However, I'd like getting this merged soon now, so that I don't need to constantly merge new changes into this branch and we can all work on the same thing.
For the two things mentioned above, I'd like to move them to separate issues.

urielsalis:
Im up to separate this PR in multiple parts too to try to reduce the changes.

Tests we can put it as a separate issue, same for yaml/toml

arisa.json
```diff
@@ -0,0 +1,122 @@
...
+        "removalReason": "Ticket has been triaged."
+      },
+
+      "futureVersion": {
+        "jql": "affectedVersion in unreleasedVersions() AND updated > -1d",
+        "message": "{panel:borderColor=orange}(!) Please do not mark _Unreleased Versions_ as affected. You don't have access to them yet.{panel}\n~{color:#888}-- I am a bot. This action was performed automagically! Please report any issues in [Discord|https://discordapp.com/invite/rpCyfKV] or [Reddit|https://www.reddit.com/r/Mojira/]{color}~"
+      },
+
+      "chk": {
+        "whitelist": ["MC", "MCL", "MCPE"],
```
urielsalis, line 88: Add MCTEST?

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,333 @@
...
+        val issues = queryCache[combinedJql] ?: jiraClient
+            .searchIssues(combinedJql)
+            .issues
+            .map { jiraClient.getIssue(it.key, "*all", "changelog") } // Get issues again to retrieve all fields
+            .filter { issue ->
+                // Ignore issues where last action was a resolve
+                val latestChange = issue.changeLog.entries.lastOrNull()
+
+                latestChange == null || // There is actually a entry
+                    !latestChange.items.any { it.field == "resolution" } || // It was a transition
```
urielsalis, line 86: [minor] Maybe all of this should be methods instead of comments 😅

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,333 @@
...
+            }
+        queryCache[combinedJql] = issues
+
+        issues
+            .map { it.key to executeModule(it) }
+            .forEach { (issue, response) ->
+                response.second.fold({
+                    processedIssues.putIfAbsent(issue, false)
+                    when (it) {
+                        is OperationNotNeededModuleResponse -> log.info("[RESPONSE] [$issue] [${response.first}] Operation not needed")
```
urielsalis, line 98: I would remove the logs for OperationNotNeeded, its too spammy :D

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,333 @@
...
+                }, {
+                    processedIssues[issue] = true
+                    log.info("[RESPONSE] [$issue] [${response.first}] Successful")
+                })
+            }
+    }
+
+    fun execute() {
+        // Cache issues returned by a query to avoid searching the same query for different modules
+        val queryCache = mutableMapOf<String, List<Issue>>()
```
urielsalis, line 112: [minor] I would open a issue to move all the cache stuff to its own class just to clean the executors. Will do in my own branch after this is merged :D

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,333 @@
...
+                    { key -> addComment(issue, config[Arisa.Modules.Crash.duplicateMessage].replace("{DUPLICATE}", key)) }
+                )
+            )
+        }
+        exec(Arisa.Modules.Empty) { issue ->
+            "Empty" to emptyModule(
+                EmptyModule.Request(
+                    issue.attachments.size,
+                    issue.description,
+                    issue.getField("environment") as? String?,
```
urielsalis, line 158: [minor] We do the issue.getField and converted to string a lot, might make sense to have a fun Issue.getFieldAsString("fieldName") that does this cast to nullable string automatically

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -9,191 +9,73 @@ object Arisa : ConfigSpec() {
...
-            )
+            val message by optional("", description = "The message that is posted when this module succeeds.")
+            val piracySignatures by optional(emptyList<String>(), description = "Signatures that indicate a pirated version of Minecraft. Default is no signatures.")
         }
 
         object RemoveTriagedMeqs : ModuleConfigSpec() {
-            val meqsTags by optional(listOf("MEQS_WAI", "MEQS_WONTFIX"))
-            val removalReason by optional("Ticket has been triaged.")
+            val meqsTags by optional(emptyList<String>(), description = "List of tags that should be removed by the bot when an issue is triaged.")
+            val removalReason by optional<String?>(null, description = "Reason Arisa should add to the edited comment for removing the tag. Default is no reason.")
```
urielsalis, line 47: I wouldnt have null until we need it in any message, it seems more unclean(we try to avoid null as much as possible)

arisa.json
NeunEinser: MCTEST has no confirmation status and no chk. Do you want to have it enabled anyways?

violine1101:
Yeah, I didn't realize that the config was already saved in JSON by the config spec. We can still move over to TOML or YAML in another PR, maybe I'll take a look at that myself. But what exactly were the issues with YAML?

This PR should probably be merged ASAP, so that other modules can be developed with this new architecture as a basis, since this change modifies basically every existing file.

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -9,191 +9,73 @@ object Arisa : ConfigSpec() {
...
-                    CrashDupeConfig("minecraft", "Failed to check session lock, aborting", "MC-10167"),
-                    CrashDupeConfig("minecraft", "Maybe try a lowerresolution texturepack", "MC-29565"),
-                    CrashDupeConfig("minecraft", "java\\.lang\\.OutOfMemoryError\\: Java heap space", "MC-12949"),
-                    CrashDupeConfig("minecraft", "try a lowerresolution", "MC-29565"),
-                    CrashDupeConfig("java", "ig[0-9]{1,2}icd[0-9]{2}\\.dll", "MC-32606")
-                )
-            )
+            val maxAttachmentAge by optional(0, description = "Max age in days the attachment can have to be considered")
+            val crashExtensions by optional(emptyList<String>(), description = "File extensions that should be checked for crash reports.")
+            val duplicateMessage by optional("", description = "Message to be send when resolving a duplicate. {DUPLICATE} will be replaced by the ticket key")
```
urielsalis, line 69: Instead of {DUPLICATE} you can use %s

Then on the usage code you do duplicateMessage.format(ticketKey) and you get the final string :D

src/main/kotlin/io/github/mojira/arisa/modules/RemoveTriagedMeqsModule.kt
```diff
@@ -4,38 +4,44 @@ import arrow.core.Either
...
 
-    private fun removeMeqsTags(comment: Comment, meqsTags: List<String>): String {
+    private fun removeMeqsTags(comment: String): String {
         val regex = (
             "MEQS(" +
             meqsTags.joinToString("|") { it.replace("MEQS", "") } +
             ")"
         ).toRegex()
-        return regex.replace(comment.body) { "MEQS_ARISA_REMOVED${it.groupValues[1]} Removal Reason: $removalReason" }
+        return regex.replace(comment) { "MEQS_ARISA_REMOVED${it.groupValues[1]}${ if (removalReason != null) " Removal Reason: $removalReason" else "" }" }
```
urielsalis, line 44: The not having it as null would make things like this nicer :D

urielsalis:
> Yeah, I didn't realize that the config was already saved in JSON by the config spec. We can still move over to TOML or YAML in another PR, maybe I'll take a look at that myself. But what exactly were the issues with YAML?

we had issues with the lists, as it was not reading them correctly

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
NeunEinser: null is the default value. If no reason is set in the config, it would be weird to have "Removal Reason: <blank>", so when it's null the module does not post `Removal Reason`

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
NeunEinser: I think I copied that from the js version when I added the crash module :D

src/main/kotlin/io/github/mojira/arisa/modules/RemoveTriagedMeqsModule.kt
NeunEinser: Well then we would have `Removal Reason: <blank>` when non is specified 😉

NeunEinser: I can set it to that if you want, just seemed a little odd to me

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
NeunEinser: ¯\_(ツ)_/¯

arisa.json
urielsalis: I do, disabling 😅

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
urielsalis: Yeah, I would make it required is what I mean

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
urielsalis: #77

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
urielsalis: #78

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -0,0 +1,334 @@
...
+
+        val issues = queryCache[combinedJql] ?: jiraClient
+            .searchIssues(combinedJql)
+            .issues
+            .map { jiraClient.getIssue(it.key, "*all", "changelog") } // Get issues again to retrieve all fields
+            .filter { issue ->
+                // Ignore issues where last action was a resolve
+                val latestChange = issue.changeLog.entries.lastOrNull()
+
+                latestChange == null || // There is actually a entry
```
urielsalis, line 85: This should be != as far as I know

NeunEinser: If there is no latest change, the filter method needs to return true.
These are exactly the inverse of what used to be in ArisaMain

urielsalis:
I was going from more general to more specific (its the usual pattern). I can revert that part if you like

src/main/kotlin/io/github/mojira/arisa/infrastructure/Cache.kt
```diff
@@ -0,0 +1,37 @@
+package io.github.mojira.arisa.infrastructure
+
+import net.rcarz.jiraclient.Issue
+import java.util.Timer
+import kotlin.concurrent.schedule
+
+class Cache {
+    private val ticketCache = mutableListOf<String>()
+    private val ticketTimer = Timer("RemoveCachedTicket", true)
+    val queryCache = mutableMapOf<String, List<Issue>>()
```
NeunEinser, line 10: Is there any point where you clear this?
This used to be a local variable in execute(), so it used to have a clear queryCache per module

PR #74, opened 2020-04-09
Tweak bot footer slightly to make it more readable

violine1101:
This is a very minor change that just tweaks the bot footer to make it a bit more readable, and in general look a bit nicer. Here's an example:

Before:
grafik

After:
grafik

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
@@ -46,10 +46,10 @@ object Arisa : ConfigSpec() {
                 "You are currently using a *non-authorized* version of Minecraft. If you wish to purchase the full game, please visit the [Minecraft Store|https://www.minecraft.net/store/minecraft-java-edition].\n" +
                 "We will not provide support for pirated versions of the game, these versions are modified and may contain malware.\n" +
                 "\n" +
-                "\n" +
                 "*Quick Links*:\n" +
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -40,16 +40,18 @@ object Arisa : ConfigSpec() {
...
+                However, this issue is {color:#FF5722}*Invalid*{color}.
+                
+                You are currently using a *non-authorized* version of Minecraft. If you wish to purchase the full game, please visit the [Minecraft Store|https://www.minecraft.net/store/minecraft-java-edition].
+                We will not provide support for pirated versions of the game, these versions are modified and may contain malware.
+                
+                *Quick Links*:
+                📓 [Issue Guidelines|https://bugs.mojang.com/projects/MC/summary] -- 💬 [Community Support|https://discord.gg/58Sxm23] -- 📧 [Customer Support|https://help.minecraft.net/hc/en-us/requests/new] -- 📖 [Game Wiki|https://minecraft.gamepedia.com/Minecraft_Wiki]
+                
+                ~{color:#888}-- I am a bot. This action was performed automagically! Please report any issues in [Discord|https://discordapp.com/invite/rpCyfKV] or [Reddit|https://www.reddit.com/r/Mojira/]{color}~
+                """.trimIndent()
```
NeunEinser, line 54: Uh that's cool syntax

violine1101: It really is!

violine1101:
> Tbh, I don't know if a different color is even needed at all, the slightly smaller text caused by `~...~` is already good enough imo.

Yeah, I figured it should probably still be separated from the comment a bit more. The main objectives of this change was to make the contrast between text and background higher, and also to have the links blend in a bit more. Before they stood out like a sore thumb.

PR #76, opened 2020-04-10
Fix #75

### LINKED ISSUE #75: RevokeConfirmation module revokes confirmation set by former volunteers
Currently, the module will revoke any confirmation which was last set by someone who is not in group helper, global-moderators or staff.
This has the unwanted effect that it will also revoke tickets confirmed by people who used to, but no longer are volunteers.

The easiest way to fix this would probably be to assume a change was made by a volunteer for old entries.

violine1101:
With this change, old confirmation status changes (at least one day ago) are treated as voluteer changes and are never reverted by the `RevokeConfirmation` module.


NeunEinser:
Okay, someone was faster haha

PR #89, opened 2020-04-10
Create issue templates

urielsalis:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
```
NeunEinser, line 6: Maybe not include the empty ones?

urielsalis: Its autogenerated from github. This are things that are added to the title/assignee if present. Not sure if it works if they are empty

NeunEinser: Ah so this will be autofilled?

Issue #38, opened 2020-04-06
Automatically re-resolve old AR tickets as Incomplete


violine1101:
It would be a great feature if Arisa would automatically close AR tickets that haven't received an update since a long amount of time (say, one or two years) with an adequate message such as
> This issue has been resolved as **Incomplete** since we didn't get an answer in a reasonable amount of time (1 year or more). If this issue still happens, please check if there already is a newer ticket about this, and if there is not, please create a new one.

This will prevent users from resurrecting old reports that have been unattended for a long time by commenting on the ticket.

This may be a bit complicated to implement since the bot currently only checks for recently updated tickets. It would probably need to be adjusted to check a filter with only old AR tickets in regular intervals.

urielsalis:
I was thinking of a possibility to have modules with custom queries, set in configuration. This would allow this kind of work

We do have to make sure we are not doing the same queries multiple times

NeunEinser:
Yes, I also thought already it would be useful to allow for custom queries per module.
That could also have the advantage that arisa will do everything missed after a downtime or when adding a new module.

I think #25 should be worked on beforehand or in conjunction with that though.

urielsalis:
!depends on #25

NeunEinser:
!depends on #59

Marcono1234:
Could we please discuss this with all moderators again or look at previous discussions? @rsh1's and my worries were that this way reports which were once valid get thrown in the same pool of reports which were never valid in the beginning because for them necessary information was never provided. So if (for whatever reason) we wanted to look at those old reports again, we might have a hard(er) time finding them. Additionally recent clean up of old reports showed that quite a large amount of them were still valid.

I know that users tend to comment on old reports, but for me it appeared this happened mostly on already resolved (non-AR) ones. Is them commenting on old AR reports really such a big issue?

violine1101:
Yes, this should definitely be discussed with all the moderators before it is implemented.

The issue with users commenting on old AR tickets is that we won't ever really notice it. Those tickets are old, so they usually don't really show up in our queues (except for the activityfeed, but you know how unreliable it is). This is why I put "create a new ticket" in my proposed message above. If the ticket is reposted, we will get some attention to the issue again.

And at the end of the day the question is, what is a dormant AR ticket really worth if it has not been updated in years? Sure it would be good to take another look at them, but there are just so many of those that manually doing that is not really feasible.

[...]

PR #91, opened 2020-04-11
Fix #86

### LINKED ISSUE #86: Use projects as default and allow modules to have their own projects via whitelist

NeunEinser:
Nice, I got pr #91 :D

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -231,7 +231,7 @@ class ModuleExecutor(
         exec(Arisa.Modules.ReopenAwaiting) { issue ->
             "ReopenAwaiting" to reopenAwaitingModule(
                 ReopenAwaitingModule.Request(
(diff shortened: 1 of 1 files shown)


NeunEinser:
I was thinking about that and an earlier version of the recent refactor actually had that. Same for resolutions btw, as the attachment module runs on all resolutions anyways

There are a couple things to consider:
- The updated > lastRun gets currently only added on modules without a custom jql.
- For example the query used for triaged meqs, without adding resolutions, would currently constantly return 40 tickets that are marked as resolved, while it's expected that the amount of stuff that happened since the last run is relatively low.

So, there are some solutions to that:
# Leave it like in this PR ^^
# Add resolutions and projects manually to the jql via the config, where a custom jql is in place, and use whitelist and resolutions only for tickets that don't have a custom jql
# We could also apply `updated > %s` to all queries by default, but that kinda defeats the point, if you only look at recent tickets anyways. Also it wouldn't work for #38

urielsalis:
Yeah, I meant projects, not resolutions. And only for the modules without a custom jql. We can do in a future PR

PR #97, opened 2020-04-11
Filter projects after caching

urielsalis:
## Purpose
Filter projects after caching the issues to avoid having to call jira multiple times

## Approach
Cache without filtering by projects, then filter on each module

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -296,7 +296,7 @@ class ModuleExecutor(
 
         issues
             .filter { it.project.key in projects }
-            .filter { it.resolution?.name ?: "unresolved" in resolutions }
+            .filter { it.resolution?.name?.toLowerCase() ?: "unresolved" in resolutions }
             .map { it.key to executeModule(it) }
             .forEach { (issue, response) ->
                 response.second.fold({
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -282,21 +282,21 @@ class ModuleExecutor(
         executeModule: (Issue) -> Pair<String, Either<ModuleError, ModuleResponse>>
     ) {
         val projects = (config[moduleConfig.whitelist] ?: config[Arisa.Issues.projects])
-        val resolutions = config[moduleConfig.resolutions].joinToString(",") { "\"$it\"" }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -282,21 +282,21 @@ class ModuleExecutor(
...
             .issues
             .map { jiraClient.getIssue(it.key, "*all", "changelog") } // Get issues again to retrieve all fields
             .filter(::lastActionWasAResolve)
 
-        cache.addQuery(combinedJql, issues)
+        cache.addQuery(jql, issues)
 
         issues
+            .filter { it.project.key in projects }
+            .filter { it.resolution?.name ?: "Unresolved" in resolutions }
```
urielsalis, line 299: You need to do toLowerCase here too, same as in line 285

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -282,21 +282,21 @@ class ModuleExecutor(
         executeModule: (Issue) -> Pair<String, Either<ModuleError, ModuleResponse>>
     ) {
         val projects = (config[moduleConfig.whitelist] ?: config[Arisa.Issues.projects])
-            .joinToString(",")
-        val resolutions = config[moduleConfig.resolutions].joinToString(",") { "\"$it\"" }
+        val resolutions = config[moduleConfig.resolutions].map { it.toLowerCase() }
```
urielsalis, line 285: I think you can do config[moduleConfig.resolutions].map(toLowerCase)

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
NeunEinser: Oh didn't I? Good catch

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -296,7 +296,7 @@ class ModuleExecutor(
 
         issues
             .filter { it.project.key in projects }
-            .filter { it.resolution?.name ?: "Unresolved" in resolutions }
+            .filter { it.resolution?.name ?: "unresolved" in resolutions }
```
urielsalis, line 299: dont you need a .toLowerCase() in the name too?

NeunEinser: Ahh, I'm doing too many things at once xD
Thank you, though 😉

PR #98, opened 2020-04-11
Resolve trash to invalid automagically

### LINKED ISSUE #58: Resolve TRASHed tickets automatically
https://github.com/mojira/arisa/blob/master/modules/trash.rb

urielsalis:
## Purpose
Resolve TRASHed tickets automatically to invalid if not done so
Fixes #58

## Approach
Filters to only TRASH and to only Unresolved in arisa.json, does a double check that its a TRASH ticket inside the module and then closes as invalid

#### Checklist
- [x] Included tests
- [x] Tested in TRASH-25123

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/ResolveTrashModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/ResolveTrashModule.kt
@@ -1,0 +1,18 @@
+package io.github.mojira.arisa.modules
+
+import arrow.core.Either
+import arrow.core.extensions.fx
+
+class ResolveTrashModule() : Module<ResolveTrashModule.Request> {
+    data class Request(
+        val projectKey: String,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/ResolveTrashModule.kt
```diff
@@ -0,0 +1,18 @@
...
+
+class ResolveTrashModule() : Module<ResolveTrashModule.Request> {
+    data class Request(
+        val projectKey: String,
+        val resolveAsInvalid: () -> Either<Throwable, Unit>
+    )
+
+    override fun invoke(request: Request): Either<ModuleError, ModuleResponse> = with(request) {
+        Either.fx {
+            assertEquals(projectKey, "TRASH").bind()
```
NeunEinser, line 14: I wonder if it would make sense to move the whitelist to the modules and add an assertIsWhitelisted or something.

PR #101, opened 2020-04-11
New crash module

urielsalis:
## Purpose
Separate crash handling logic to new library

## Approach
Still old code, adapted to use new lib. Needs to be refactored

UNTESTED

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -40,28 +43,64 @@ class CrashModule(
...
+                .filter { it.second.isRight() }
+                .map { extractCrash(it) }
+                .filter { it.second is Crash.Minecraft || it.second is Crash.Java }
+                .toList()
 
-            val mostRelevantInfo = infos.reduce(::getMoreRelevantInfo)
+            assertNotEmpty(crashes).bind()
 
-            if (mostRelevantInfo.modded) {
+            if (crashes.any { it.second is Crash.Minecraft && (it.second as Crash.Minecraft).modded }) {
```
NeunEinser, line 57: The current module prefers unmodded crashes, which makes sense, as one person might attach a modded crash, even though the mai report is about a vanilla crash.

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -40,28 +43,64 @@ class CrashModule(
...
                 addModdedComment().toFailedModuleEither().bind()
                 resolveAsInvalid().toFailedModuleEither().bind()
             } else {
-                val configs = crashDupeConfigs.filter(::isConfigValid)
-                val key = getDuplicateKey(mostRelevantInfo, configs)
-                assertNotNull(key).bind()
+                val sortedMap = crashes.toMap().toSortedMap(compareByDescending { it.created })
+                val key = findDuplicate(sortedMap, crashDupeConfigs)
```
NeunEinser, line 62: I'm also not too sure it should try to find a duplicate in all crashes.
The current module only sees the most recent crash that is not above a certain age. This might be fine if the max attachment age is less then 30 days though (as supposed to the original js module). I'd say it would be good if it just used the lastRun time.

It might also be good to ignore tickets with a confirmation status, Mojang Priority, ADO or assignee, just in case someone attaches an irrelevant crash report to an otherwise valid ticket.

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: So how is the logic supposed to be for modded crashes?

1 crash report and its modded -> invalid
2 crash reports, one its modded -> scan that one
2 crash reports, both modded -> invalid?

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: This looks for any match in less than 30 days. I was thinking on the case that a user posts multiple logs(like launcher log and multiple crashes).

Agree with checking confirmation status + Mojang priority (not sure how you would check ADO)

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
NeunEinser: > 2 crash reports, one its modded -> scan that one

Not sure which one you meant with "scan that one", but the current module would scan the unmodded one in that case, which makes sense imo.

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
NeunEinser: ADO is probably another custom field that would need a config entry.

urielsalis: It would still prefer the most recent crash of all

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: the test for "should prefer crash that is not modded, if modded crash appears first" seems to be passing

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: Added the confirmed and mojang priority check

urielsalis: I would exclude the crash module from MCPE anyways

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: scan the other, yes

urielsalis: Made it so it scans all crashes, but if there is no match and crash is modded to close as invalid

(all the detections we have happen whenever its modded or not, so its still useful to close them as duplicate instead of just invalid which would make the user create a new ticket just to get closed as duplicate)

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: This only operates in MC, MCL and MCTEST, we shouldnt need ADO.

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -95,7 +95,7 @@ class ModuleExecutor(
             "CHK" to chkModule(
                 CHKModule.Request(
                     issue.getFieldAsString(config[Arisa.CustomFields.chkField]),
-                    ((issue.getField(config[Arisa.CustomFields.confirmationField])) as? JSONObject)?.get("value") as? String?,
+                    issue.getCustomField(config[Arisa.CustomFields.confirmationField]),
```
NeunEinser, line 98: Nice!

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
NeunEinser: Oh yes, good point

Issue #60, opened 2020-04-07
Automatically detect tickets in other languages


violine1101:
This may be difficult to implement, but a great feature would be if Arisa would automatically detect if a ticket has been reported in a language other than English. For languages which don't use a Latin script this is easy to do (just check whether 90% or more of the characters in the ticket are non-ASCII), but there also exist Java libraries that provide the option to determine what language a text is from.

Either Arisa would just close these tickets as Invalid (easier to implement) or it would use machine translation in order to translate the ticket (difficult to implement).

urielsalis:
https://github.com/pemistahl/lingua can be used, with the top 10 languages spoken (to make sure memory usage keeps down and accuracy is high)

violine1101:
Thanks for that link! That seems like the way to go. I'm not exactly sure yet what exactly Arisa should do when it detects a non-English or an unknown language. I guess Arisa could post a message that has been translated to the detected language if a known non-English language is detected. But what if it's not one in our list? Just assume that it's not English or leave it open for a human to check?

urielsalis:
I would say if its more than 90% confident to just close as invalid

urielsalis:
Do note we have to include english for that

urielsalis:
and not translate, thats not useful 😓

Pokechu22:
I think it would be useful for the closure comment to be included in both the detected language and English, if we can reliably do such translation.

The other thing to be mindful of is tickets that have foreign characters (e.g. font issues including [MC-148898](https://bugs.mojang.com/browse/MC-148898 "[MC-148898] Hebrew letters aren't assigned to their correct textures")), and translation issues on the WEB project (where they're valid, unlike with MC).  Possibly that can just be solved by only resolving them on first creation, and if they're reopened then they won't be resolved again (since I don't think it's likely that a user will edit a ticket that was originally in English to be in some other language).

violine1101:
Yes, that's something I'd need to check, I'm not sure if lingua handles loanwords correctly. If it doesn't, I'd probably need to split the text up into multiple different parts, analyze them separately and only trigger the bot if more than 80% or so of these parts are not in English according to lingua. But as far as I can tell, lingua doesn't simply only check the characters in the letters, it seems to take a more sophisticated approach.

[...]

PR #104, opened 2020-04-12
Implement first working test version of language detection module

violine1101:
## Purpose
This change adds a module that automatically detects whenever a ticket is not in English as per #60. It does not resolve the ticket, it only leaves a restricted comment that explains what it would have done, in order to test how this module would work and how accurate it is in practice.

## Approach
I used [`lingua`](https://github.com/pemistahl/lingua/) for this module. Lingua chooses the language it thinks is most likely correct for us from a list of languages. Currently, this list of languages is:
* Chinese
* English
* French
* German
* Japanese
* Korean
* Polish
* Portuguese
* Russian
* Spanish

If the resulting language is unknown (i.e. not on this list or unclear), or if it is English, no action is taken. However, if it is any other language, the bot leaves a restricted comment detailing which change it would have done to the ticket.

## Future work

There are multiple issues with this current setup:
1. The bot incorrectly closes tickets that are half in English and half in another language. This sometimes happens if someone who is not familiar with English uses a machine translator to translate the text, but includes the original as well. These tickets should be considered valid. Currently, `lingua` only returns the language it thinks is most likely to be correct, but it doesn't return any certainty values or results for other languages.
I implemented the workaround proposed by @Marcono1234 in https://github.com/mojira/arisa-kt/issues/60#issuecomment-612653038, however it causes major performance issues since the language detector would need to be rebuilt every time a ticket that potentially is in another language is examined. Here is a patch file with this workaround: https://gist.github.com/violine1101/58f23fb0d220ec6ac1ef0bc0cda446a0
It can probably be improved by storing the language detectors somewhere, but that would cause higher memory consumption.
However, it would obviously be optimal if `lingua` would already give us the certainty values by itself, so that we don't need to use workarounds like that.
2. Tickets in a language that is not included in the list above are likely to be incorrectly identified. If they're not closed by Arisa (which would be fine), they're closed because Arisa thinks they're in another language, which is not necessarily the correct one, and could cause confusion. This could be mitigated by adding more languages to the list, which in turn would cause more memory consumption.
3. It is unclear how memory-intensive this change is. I tried to measure it, however I don't really have a reliable method to do this. It's likely to be rather resource-intensive, but I'm not sure about how or if it is possible to use less RAM.

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-70 and MCTEST-71


src/main/kotlin/io/github/mojira/arisa/modules/LanguageModule.kt
```diff
@@ -0,0 +1,58 @@
...
+
+class LanguageModule(
+    private val allowedLanguages: List<Language> = listOf(
+        Language.ENGLISH,
+        Language.UNKNOWN
+    )
+) : Module<LanguageModule.Request> {
+
+    companion object {
+        private val detector = LanguageDetectorBuilder.fromLanguages(
```
urielsalis, line 18: I would move this out of the companion object and just have it in the class. Companion objects have some problems(they generate way to much code) and this doesnt really need to be static

Modules are initialized only once anyway

src/main/kotlin/io/github/mojira/arisa/modules/LanguageModule.kt
```diff
@@ -0,0 +1,58 @@
...
+            assertLanguageIsNotAllowed(allowedLanguages, detectedLanguage).bind()
+
+            addLanguageComment(detectedLanguage).toFailedModuleEither().bind()
+            resolveAsInvalid().toFailedModuleEither().bind()
+        }
+    }
+
+    private fun getDetectedLanguage(summary: String?, description: String?): Language {
+        return detector.detectLanguageOf("${summary ?: ""} ${description ?: ""}")
+    }
```
NeunEinser, line 52: I think it would make sense to not have this function in the actual module.
For jira lib stuff, we have `Jira.kt`, which makes it quite convenient for the module, especially when testing. Some of your current tests try to test the accuracy of the module instead, which in my opinion doesn't belong there.

I'd rather have tests that we can actually control, and test the accuracy of the lib on the tracker with hidden comments instead.

PR #106, opened 2020-04-12
Language module via api

urielsalis:
## Purpose
Detect the language of a ticket and close it if we are confident by more than 80% that is not english

## Approach
Branched off from #104, but using a API instead of lingua

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


build.gradle.kts
```diff
@@ -9,6 +9,7 @@ version = "1.0-SNAPSHOT"
 
 repositories {
     mavenCentral()
+    maven("https://dl.bintray.com/cbeust/maven")
```
Marcono1234, line 12: Might be better to move that below JitPack. I am not completely sure if there are any restrictions on private Bintray Maven repositories, but if that is not the case the owner of that repository could (for whatever reason) publish modified 3rd party artifacts which we are currently retrieving from JitPack.

Marcono1234:
Will this module run for private reports as well? If so I think we should exclude them because we would be sending sensitive information to a 3rd party server otherwise.

urielsalis:
Not anymore

Issue #51, opened 2020-04-07
Automatic updating linked field


urielsalis:
When marking a report as duplicate of another, either counting the number of duplicates on the parent and setting the pinned field, or incrementing the pinned field if counting isn't possible or is less efficient.

Moved from https://github.com/mojira/arisa-js/issues/5

urielsalis:
Do we still need it?

NeunEinser:
When I became helper, pinned and linked fields already weren't used anymore.

What were those fields intended for? Just a counter seems odd, as Jira already has that.

urielsalis:
I agree. Closing

NeunEinser:
Reopening for further discussion, see #57

violine1101:
This would be a very useful statistic to see which tickets are frequently reported (i.e. if a recently reported ticket has a large number of duplicates, it's likely to be a severe issue). I don't think anything would be wrong with reviving the Pinned or Linked field (personally I'd prefer if the Linked field were to be used, as it's more appropriately named).

As far as I'm aware, the Pinned and Linked fields stopped being used before Arisa was developed because it was tedious to update them (it needed to be done manually as far as I know). This is also why this issue already existed in the original Arisa repository.

Of course this would need to be discussed with Mojang & the other mods as well though, but seeing as these fields are currently unused, I don't think anything speaks against giving them a purpose.

Marcono1234:
`Linked` has been used differently in some cases where it contains the report ID instead of the count, see JQL: `Linked is not empty order by Linked DESC`
Though it appears this only affects a few older reports, however it would require clean up of them then.

We might also have to think about how trashed duplicates or update duplicated links (e.g. because the link was incorrect) should be handled, see for example MC-297 which currently only has 7 duplicates (but actually had multiple hundreds).

violine1101:
I think it'd probably be most useful if it always contains the current number of duplicates, and that is probably also the easiest way to do it.

PR #107, opened 2020-04-13
add linked module

NeunEinser:
## Purpose
Adds a module that automatically updates the linked (#51)

## Approach
Adds a new very cool module

#### Checklist
- [x] Included tests
- [x] Tested in [MC-163575](https://bugs.mojang.com/browse/MC-163575)

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -17,6 +17,7 @@
       "checkInterval": 10
     },
     "customFields": {
+      "linked": "customfield_11100",
       "chkField": "customfield_10701",
       "confirmationField": "customfield_10500",
       "mojangPriorityField": "customfield_12200",
--- a/arisa.json
+++ b/arisa.json
@@ -218,6 +218,13 @@
         "whitelist": [
           "TRASH"
         ]
+      },
+      "updateLinked": {
+        "whitelist": [
+          "MC",
+          "MCL",
+          "MCPE"
+        ]
       }
     }
   }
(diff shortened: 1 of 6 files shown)


NeunEinser:
I am currently running a modified version of this locally which gradually updates all open reports with this.

arisa.json
```diff
@@ -1,223 +1,231 @@
-{
```
urielsalis, line 1: I think you have a different config for json formatting

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -50,29 +52,30 @@ class ModuleExecutor(
     private val config: Config,
     private val cache: Cache
 ) {
-    private val attachmentModule: AttachmentModule =
+    private val attachmentModule =
```
urielsalis, line 55: Actually seems like all autoformat rules are a bit wrong :/

NeunEinser: Did you leave those intentionally?
I removed them on purpose with IntelliJ, since I see no value in `val attachmentModule: AttachmentModule = AttchamentModule(...`

arisa.json
NeunEinser: I think this might be because it was CRLF before and now changed to LF, like all other files.

NeunEinser: Yes, if you hide whitespace changes, you see what actually changed:
image

Marcono1234: Under Windows you can set `core.autocrlf` to `true`. This will make Git convert line endings to CRLF in the working directory and converts line endings of all added files to LF when added, see https://git-scm.com/docs/gitattributes#_end_of_line_conversion

Or we could add a `.gitattributes` file which specifies `* text=auto` so all files added are normalized (if they are detected as text) regardless of the `core.autocrlf` setting a contributor has.

NeunEinser: The file I committed is using LF now, before it was CRLF in the working dir for whatever reason.
All the other files did use LF already.

Issue #78, opened 2020-04-10
Add tests to ModuleExecutor


urielsalis:
To be done after #67

PR #125, opened 2020-04-20
ModuleRegistry

NeunEinser:
## Purpose
This pr moves all the module initializations and how to send requests out of the module executor. This makes the execute method in particular more concise and removes all the logic that is required to send requests to a module from it.

## Approach
The module executor's execute method is currently very long, as it has to specify a function per module that is called to send a request to a module for a specific issue.
With this PR the Module Executor does not know about any of the modules anymore, instead it knows about a ModuleRegistry. The Module registry contains a list, where each entry contains the config entry of the module and a function that takes an issue and sends a request to the module.

The ModuleRegistry's register method is constructed in such a way that you only have to specify relevant information now.
In the ModuleExecutor you used to call the exec method like this:
```
exec(<config>) { issue ->
    <name> to tryExecuteModule {
        <module>(
            <request-constructor>(
                ...
            )
        )
    }
}
```
While it is simplified to this in the Registry:
```
register(
    <name>,
    <config>,
    <module>
) { issue ->
    <request-constructor>(
        ...
    )
}
```
Note how `tryExecuteModule` is gone from there now, and the request is just initialized, but you don't need to specify how to send it to the module anymore.

## Future Work
With module logic being moved out of the module executor, it should be easier now to test the ModuleExecutor's logic with a mocked registry. (see #78)


urielsalis:
This is really nice, we need to do more refectors like this now that we have our structure pretty clear

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -0,0 +1,379 @@
...
+        register(
+            "RevokeConfirmation",
+            Modules.RevokeConfirmation,
+            RevokeConfirmationModule()
+        ) { issue ->
+            RevokeConfirmationModule.Request(
+                issue.getCustomField(config[CustomFields.confirmationField]),
+                issue.changeLog.entries
+                    .flatMap { e ->
+                        e.items
```
urielsalis, line 337: It would be nice to extract this kind of methods to some mapper functions and extract it to another file

NeunEinser: Yeah, that would probably be a bit more organized.
I kinda like having *everything* about the module's functions in one place, but the init function is a bit cluttered like this :D

urielsalis: feel free to merge by the way, thats  future work :D

urielsalis: I see it as a jira thing, so we should really create a jira package under infrastrucutre and put a Calls.kt and Domain.kt

PR #138, opened 2020-04-21
Add ConfirmParentModule

SPGoding:
## Purpose
Resolve #123

## Approach
Added a new ConfirmParentModule which will set the Confirmation Status of tickets whose Linked is greater than `linkedThreshold` and Confirmation Status exists in `confirmationStatusWhitelist` to the value specified in `targetConfirmationStatus`.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

PR #141, opened 2020-04-21
Filter out duplicates that are reported by the same user

### LINKED ISSUE #140: Filter out issues with the same reporter when updating the Linked field
As users will be able to confirm their own tickets by duplicating themselves after #138 is merged, we might want to filter out duplicates that are reported by the same person when updating the `Linked` field.

SPGoding:
## Purpose
Resolve #140.

## Approach
By involving the reporters of inward issues in `Request`.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -369,8 +369,9 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config) {
             UpdateLinkedModule()
         ) { issue ->
             UpdateLinkedModule.Request(
+                issue.reporter.name,
                 issue.issueLinks
-                    .map { it.type.name },
+                    .map { UpdateLinkedModule.IssueLink(it.inwardIssue.reporter.name, it.type.name) },
                 issue.getField(config[CustomFields.linked]) as? Double?,
                 ::updateLinked.partially1(issue).partially1(config[CustomFields.linked])
             )
--- a/src/main/kotlin/io/github/mojira/arisa/modules/UpdateLinkedModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/UpdateLinkedModule.kt
@@ -2,23 +2,36 @@ package io.github.mojira.arisa.modules
 
 import arrow.core.Either
 import arrow.core.extensions.fx
+import arrow.syntax.function.partially1
 
 class UpdateLinkedModule : Module<UpdateLinkedModule.Request> {
     data class Request(
-        val links: List<String>,
+        val parentReporter: String,
+        val links: List<IssueLink>,
         val linkedField: Double?,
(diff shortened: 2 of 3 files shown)


src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -369,8 +369,9 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config) {
             UpdateLinkedModule()
         ) { issue ->
             UpdateLinkedModule.Request(
+                issue.reporter.name,
                 issue.issueLinks
-                    .map { it.type.name },
+                    .map { UpdateLinkedModule.IssueLink(it.inwardIssue.reporter.name, it.type.name) },
```
NeunEinser, line 374: This will not work.
inwardIssue has only a very limited amount of fields. This is the api response I got from a ticket:
```json
"issuelinks": [
	{
		"id": "161296",
		"self": "https://bugs.mojang.com/rest/api/2/issueLink/161296",
		"type": {
			"id": "10102",
			"name": "Duplicate",
			"inward": "is duplicated by",
			"outward": "duplicates",
			"self": "https://bugs.mojang.com/rest/api/2/issueLinkType/10102"
		},
		"inwardIssue": {
			"id": "241711",
			"key": "BDS-884",
			"self": "https://bugs.mojang.com/rest/api/2/issue/241711",
			"fields": {
				"summary": "Minecarts despawning",
				"status": {
					"self": "https://bugs.mojang.com/rest/api/2/status/5",
					"description": "A resolution has been taken, and it is awaiting verification by reporter. From here issues are either reopened, or are closed.",
					"iconUrl": "https://bugs.mojang.com/images/icons/statuses/resolved.png",
					"name": "Resolved",
					"id": "5",
					"statusCategory": {
						"self": "https://bugs.mojang.com/rest/api/2/statuscategory/3",
						"id": 3,
						"key": "done",
						"colorName": "green",
						"name": "Done"
					}
				},
				"issuetype": {
					"self": "https://bugs.mojang.com/rest/api/2/issuetype/1",
					"id": "1",
					"description": "A problem which impairs or prevents the functions of the product.",
					"iconUrl": "https://bugs.mojang.com/secure/viewavatar?size=xsmall&avatarId=18903&avatarType=issuetype",
					"name": "Bug",
					"subtask": false,
					"avatarId": 18903
				}
			}
		}
	}
]
```

I'm not sure what the best solution to this is. Getting every linked issue on every ticket would generate a lot of traffic and is probably not worth the afford.
I think it would be better for #138 to receive a function that gets (and maps) an issue to a module data class and assert that one of the duplicates is not from the one with the same reporter.

SPGoding: Oops, that's too sad. I'd prefer just raise the threshold of the Linked field in #138🤦

PR #153, opened 2020-04-23
Add ReplaceCommentModule

### LINKED ISSUE #103: Remove URL prefix from issue keys
There have been multiple cases where reporters or commenters include the URL for issue keys or create a link with the issue key as name:
- `https://bugs.mojang.com/browse/MC-123`
- `[MC-123|https://bugs.mojang.com/browse/MC-123]`

This is unnecessary because Jira detects issue keys automatically and adds links and formats them, however not all users know that.
Having these as link instead of issue key is irritating because they
- don't have hover text
- aren't strike through when the issue is resolved

Maybe it would be worth having Arisa change the links?

Security considerations: In case Arisa checks if the referenced issue exists before removing the URL prefix, then it must not consider private issues, otherwise this would allow non-staff users querying which private issues exist by checking whether Arisa removes the URL prefix.

SPGoding:
## Purpose
Resolve #103.
## Approach
By adding a new `ReplaceCommentModule`.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -269,6 +269,26 @@
           "MCL",
           "MCPE"
         ]
+      },
+      "replaceComment": {
+        "resolutions": [
+          "Awaiting Response",
+          "Cannot Reproduce",
+          "Done",
+          "Duplicate",
+          "Fixed",
+          "Incomplete",
+          "Invalid",
+          "Unresolved",
+          "Won't Fix",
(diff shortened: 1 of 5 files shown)


SPGoding:
Ah, I will open a PR later to improve this module

PR #163, opened 2020-04-24
Only rerun failed tickets after a restart

NeunEinser:
## Purpose
Currently, whenever a module fails, the issue ID is saved and Arisa will run all modules again on the next run.
This causes the log to be spammed with errors whenever there is an issue a module fails on.

## Approach
This PR changes it so that Arisa will only try to re-run issues were a module failed after a restart.

#### Checklist
- [x] Tested in shadow mode


urielsalis:
so if I understand correctly we rerun failed tickets?

NeunEinser:
Currently, tickets that fail rerun on every execution.
After this, they will only be rerun at the first execution after a restart.

PR #171, opened 2020-04-29
Feature/transfer dupes

### LINKED ISSUE #145: Transfer links from duplicated reports to the parent
**What is the problem the module is aiming to solve? Please describe.**
It would be nice if Arisa could automatically relink all tickets from a duplicate to the parent.

**Describe the solution you'd like**
A cool new module with all the bells and whistles.

**Describe alternatives you've considered**
Uh?

**Additional context**
Maybe we should rework these issue presets a bit.

NeunEinser:
## Purpose
Resolves #145 

## Approach
Added a new module.
Since this module shared some code with the TransferVersionsModule, I decided to introduce a common superclass for both modules.
This means that the TransferVersionsModule is also rewritten to work with that abstract superclass.

#### Checklist
- [x] Included tests
- [x] Tested in [MCTEST-77](https://bugs.mojang.com/browse/MCTEST-77)


NeunEinser:
I'm not sure how you mean abstracting common functions.

Where would the common fuctions go, if not in an abstract module?

urielsalis:
yeah, thats why I approved, is more of a naming thing :D

NeunEinser:
If my names are bad, please tell me, I can change those easily ;)

urielsalis:
Dont have any suggestions either 😅

PR #178, opened 2020-05-02
Feature/revoke fields

### LINKED ISSUE #147: Module to revoke changes done by none-volunteers to certain fields
## The Problem
Similarily to the RevokeConfirmationModule, it would be nice to have a module that revokes changes done to certain fields, e.g. Mojang Priority.

## Possible Implementation
It would probably be best to make a more general module that lets you configure fields and permission groups and then replace the current RevokeConfirmationModule with that more general module.

NeunEinser:
## Purpose
Resolves #147 

## Approach
This replaces the RevokeConfirmationModule with a module that can be configured to revoke any field

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


arisa.json
```diff
@@ -216,6 +216,76 @@
...
+          },
+          {
+            "fieldId": "customfield_12201",
+            "fieldName": "Triaged Time",
+            "fieldType": "STRING",
+            "defaultValue": null,
+            "permissionGroups": [
+              "staff"
+            ],
+            "message": "{panel:borderColor=orange}(!) Don't try to change \"Triaged Time\". It may only be changed by the developers of the game. Consider this a warning.{panel}"
```
urielsalis, line 287: the "Consider this a warning" seems a bit direct 😅, can we use something similar to the the keep private message?

NeunEinser: This is based on `{panel:borderColor=orange}(!) Don't try to change "Mojang Priority". It may only be changed by the developers. Consider this a warning.{panel}` from `Panel - Mojang Priority`  from the helper messages.
Maybe it should be a little less harsh for bot messages or fields that only require helper?

PR #224, opened 2020-05-06
Relog if execution failed with an exception(not FailedModuleException…

### LINKED ISSUE #113: Arisa doesn't try reconnecting and crashes if bugtracker is down
**Describe the bug**
Arisa doesn't try reconnecting and crashes if bugtracker is down or the connection is refused

It should wait a minute and retry

urielsalis:
…) and more than a minute passed

## Purpose
Fix #113

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
@@ -20,12 +20,13 @@ fun main() {
         .from.env()
         .from.systemProperties()
 
-    val jiraClient =
+    var jiraClient =
         connectToJira(
             config[Arisa.Credentials.username],
             config[Arisa.Credentials.password],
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -59,6 +60,14 @@ fun main() {
 
             lastRunFile.writeText("${curRunTime.toEpochMilli()}$failed")
             lastRunTime = curRunTime
+        } else if (lastRelog.plus(1, ChronoUnit.MINUTES).isAfter(Instant.now())) {
```
NeunEinser, line 63: Maybe it would be better to do that only if the last successful run was more than a minute ago?
There are sometimes hiccups where a run fails, but the next run 10 seconds later succeeds again, without a restart or anything.

PR #232, opened 2020-05-17
Change ReopenAwaiting to assert comment was created after last run

### LINKED ISSUE #151: ReopenAwaiting may incorrectly ignore a comment because it's edited
## The Bug
https://github.com/mojira/arisa-kt/blob/e71bdfdda855e80da19dfa414ba76c0732777301/src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt#L64-L68
This assertion will prevent arisa from considering any tickets were the latest comment is edited.
However, both the creation and the edit could have happened during the last run, which is especially true after a longer downtime.

## Possible Solution
Since modules can now access the `lastRun` time now, I propose that this check is changed, so that all comments that were created after the last run are considered instead.

SPGoding:
## Purpose
Resolve #151.
## Approach
Assert the last comment was created after last run.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt
@@ -30,8 +30,8 @@ class ReopenAwaitingModule(
             val resolveTime = changeLog.last(::isAwaitingResolve)
                 .created
             val lastComment = comments.last()
-            assertGreaterThan(lastComment.created.toEpochMilli(), resolveTime.toEpochMilli()).bind()
-            assertGreaterThan(lastComment.created.toEpochMilli(), lastRun.toEpochMilli()).bind()
+            assertAfter(lastComment.created, resolveTime).bind()
+            assertAfter(lastComment.created, lastRun).bind()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt
```diff
@@ -30,9 +31,7 @@ class ReopenAwaitingModule(
                 .created
             val lastComment = comments.last()
             assertGreaterThan(lastComment.created.toEpochMilli(), resolveTime.toEpochMilli()).bind()
-            assertUpdateWasNotCausedByEditingComment(
-                updated.toEpochMilli(), lastComment.updated.toEpochMilli(), lastComment.created.toEpochMilli()
-            ).bind()
+            assertGreaterThan(lastComment.created.toEpochMilli(), lastRun.toEpochMilli()).bind()
```
NeunEinser, line 34: [Minor] Not sure why the line above doesn't use it either, but there is also `assertIsAfter` which works for Instance

PR #235, opened 2020-05-17
Use inputStream when fetching helper messages

SPGoding:
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
@@ -115,7 +115,9 @@ data class HelperMessages(
         fun deserialize(json: String) = Klaxon().parse<HelperMessages>(json)
 
         fun fetch() = with(URL(url).openConnection() as URLConnection) {
-            deserialize(content as String).rightIfNotNull { Error("Couldn't download or deserialize helper messages") }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
```diff
@@ -115,7 +115,9 @@ data class HelperMessages(
         fun deserialize(json: String) = Klaxon().parse<HelperMessages>(json)
 
         fun fetch() = with(URL(url).openConnection() as URLConnection) {
-            deserialize(content as String).rightIfNotNull { Error("Couldn't download or deserialize helper messages") }
+            Klaxon().parse<HelperMessages>(inputStream).rightIfNotNull {
```
NeunEinser, line 118: Where does the variable `inputStream` come from?

urielsalis: he is inside a with, so from the UrlConnection

SPGoding: It's a property of `URLConnection`

NeunEinser: Oh, the with, it's probably a field. merging.

PR #241, opened 2020-05-18
Fix ReopenAwaitingModule

### LINKED ISSUE #158: Do not reopen AR tickets if a staff-restricted comment contains a certain keyword
## The Problem
In rare circumstances, it might be desirable to keep Arisa from reopening a ticket and leave the decision to reopen the ticket up to a human moderator.

I'm not exactly sure how often this would be used, but it might be a nice tool to have in case an AR ticket is spammed with unrelated comments.

## Possible Solution
Before reopening an AR ticket that has been updated, check whether any staff-restricted comment contains a certain keyword (such as `ARISA_KEEP_AR` or something like that), and if there is a comment with that keyword, don't reopen the ticket.

### LINKED ISSUE #186: Arisa doesn't reopen AR tickets if the reporter updates the ticket, but doesn't comment
## The Bug
Arisa should reopen an AR ticket whenever the reporter updates it. However it does not do that if the reporter doesn't leave a comment as well.

## Example Tickets
https://bugs.mojang.com/browse/REALMS-3560
https://bugs.mojang.com/browse/MC-181471

### LINKED ISSUE #240: ReopenAwaiting doesn't reopen the ticket if a valid comment is followed by an invalid comment
## The Bug
If a valid comment is followed by an invalid (e.g. restricted to helpers) comment, Arisa will only test the last comment and thus won't reopen the ticket.
## Expected behavior
The ticket should be reopened.
## Possible Solution
Add predicate in the `last` method.

SPGoding:
## Purpose
Resolve #158; Fix #186; Fix #240.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/modules/Module.kt
```diff
@@ -47,8 +52,8 @@ fun <T> assertNotNull(e: T?) = when (e) {
     else -> Unit.right()
 }
 
-fun assertOr(vararg list: Either<OperationNotNeededModuleResponse, ModuleResponse>) =
-    if (list.any { it.isRight() }) {
+fun assertEither(vararg list: () -> Either<OperationNotNeededModuleResponse, ModuleResponse>) =
```
NeunEinser, line 55: Why was the parameter changed to a list of functions?
I don't see how any of the other changes would not work if this was kept as a list of Eithers

src/test/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModuleTest.kt
```diff
@@ -74,7 +143,7 @@ class ReopenAwaitingModuleTest : StringSpec({
 
     "should return OperationNotNeededModuleResponse when there were multiple resolves, but no comment after the last resolve." {
```
NeunEinser, line 144: This test should probably be replicated for a change

src/main/kotlin/io/github/mojira/arisa/modules/Module.kt
SPGoding: I suppose the `any` method returns `true` once it finds an element matching the predicate, so by using lambda not all asset are executed.

Marcono1234:
Added `MEQS_KEEP_AR` to the internal report about all MEQS tags.

Issue #209, opened 2020-05-06
Only map issues once per run


NeunEinser:
## The Problem
Currently, for every module, all issues will be mapped individually to internal data classes. This is, however, not strictly necessary and could be optimized to only happen once per run.

## Possible Solution
We currently already have a query cache in which all issues that were updated since the last successful run are cached, in order to reduce api requests. Instead of caching the classes that were returned by the API, it would make sense to already map the issue to internal classes in there and then just send those to the modules instead.

Issue #157, opened 2020-04-23
Add a command for adding archived versions to affected or fix versions


violine1101:
## The Problem
In some rare circumstances, it might be desirable to add an archived version to either the affected or the fix versions. This cannot be done via JIRA natively and needs to be done via the API.

## Possible Solution
When Arisa receives a comment with the contents `!fixed <version>` (where `<version>` is any version available on JIRA), it will add the respective version to the "Fix version/s" list (and resolve the ticket as fixed if it isn't already). Similarily, when it receives a comment containing `!affects <version>` it will add the version to the list of affected versions.

Since this action is irreversible, the command should only be able to moderators, and maybe helpers as well. Commands by other users should be ignored.

It might be a good idea to restrict the comment as soon as it is processed (if it isn't restricted already), so that normal users don't spam the comments with those commands.

Of course the commands could also look differently, this is just my suggestion.

Marcono1234:
Re #158: Maybe (?) use a command format which is consistent for both actions and indicates that this is related to Arisa, so we later know the meaning of the comment.

violine1101:
Would you have any suggestions? I'm not really sure about the command format I proposed myself, but the syntax should probably not be more complicated than it needs to be.

NeunEinser:
If #158 is supposed to have the same format, we need to have a tag format that can be detected and removed, as in RemoveNonStaffMeqs.
Otherwise we face the problem of regular users being able to use those tags and no good way of telling whether it's a former volunteer or someone who posted it and got hidden by a moderator.

NeunEinser:
If #158 is supposed to have the same format, we need to have a tag format that can be detected and removed, as in RemoveNonStaffMeqs.
Otherwise we face the problem of regular users being able to use those tags and no good way of telling whether it's a former volunteer or someone who posted it and got hidden by a moderator.

violine1101:
It should probably only act on incoming comments, there's not really a point of checking older comments as this would be a new thing. Comments like this that don't come from team members should probably just be ignored.

NeunEinser:
There is no way to distinguish between a comment from a former volunteer and the comment from an actual volunteer.
While it would obviously be an option to ignore older comments for adding archived versions, this does not work for #158, and Marcono proposed that both #158 and this should be consistent.

violine1101:
[...]

PR #244, opened 2020-05-18
Fix tranferring old affected versions

### LINKED ISSUE #229: Don't copy versions older than the oldest version of the duplicated report
## The Problem
TransferVersionsModule currently copies all affected versions to the duplicated report. This can cause problems when a report has a version set as affected despite the bug not being present in that version. If a staff member then fails to notice that (which is not unlikely), that version will be erroneously added to the duplicated report.
This is especially problematic for Bedrock edition due to the incorrect behavior of the FutureVersionModule described in #228. And in general it happens now and then that users just select whatever version is offered, including the latest release, despite a bug not being present there.

## Possible Solution
To prevent these cases it might therefore be good to only copy the affected versions which are newer than the oldest affected version of the duplicated report. This is not completely fool proof because it can happen (as it was the case for 1.8.8) that a release is published after snapshots for the next release have been published. However, this happens rarely and should therefore not be a too big issue.

Example:
1. User creates MC-456, sets 1.15.2 (despite not being affected) and 20w19a
2. Staff member resolves it as duplicate of MC-123 (current affected versions: 20w18a)

Currently Arisa would add 1.15.2 and 20w19a to MC-123, despite 1.15.2 not being affected.
With the proposed change it would only add 20w19a, but not 1.15.2 because it was released before the oldest affected version of the duplicated report (i.e. 20w18a).

SPGoding:
## Purpose
Fix #229.
## Approach
By checking the `releaseDate` of versions.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -4,6 +4,7 @@ import arrow.core.Either
 import arrow.core.left
 import arrow.core.right
 import arrow.syntax.function.partially1
+import arrow.syntax.function.partially2
 import arrow.syntax.function.pipe
 import arrow.syntax.function.pipe2
 import arrow.syntax.function.pipe3
@@ -107,7 +108,7 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
         requestCreator: (Issue, Instant) -> T
     ) = { issue: Issue, lastRun: Instant ->
         config::class.simpleName!! to
-                ({ lastRun pipe (issue pipe2 requestCreator) pipe module::invoke } pipe ::tryExecuteModule)
+            ({ lastRun pipe (issue pipe2 requestCreator) pipe module::invoke } pipe ::tryExecuteModule)
     } pipe (getJql pipe2 (config pipe3 ModuleRegistry::Entry)) pipe modules::add
 
     @Suppress("TooGenericExceptionCaught")
@@ -229,8 +230,8 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
         register(Modules.TransferVersions, TransferVersionsModule()) { issue ->
             AbstractTransferFieldModule.Request(
                 issue.key,
-                issue.getLinks(jiraClient, ::addAffectedVersionById, ::getVersionsGetField),
-                issue.versions.map { it.id }
+                issue.getLinks(jiraClient, ::addAffectedVersionById, ::getVersionsGetField.partially2 { Unit.right() }),
+                issue.getVersions { Unit.right() }
             )
         }
 
@@ -368,7 +369,7 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
                 val intervalStart = now.minus(config[Modules.UpdateLinked.updateInterval], ChronoUnit.HOURS)
                 val intervalEnd = intervalStart.minusMillis(now.toEpochMilli() - lastRun.toEpochMilli())
                 return@register "updated > ${lastRun.toEpochMilli()} OR (updated < ${intervalStart.toEpochMilli()}" +
-                        " AND updated > ${intervalEnd.toEpochMilli()})"
+                    " AND updated > ${intervalEnd.toEpochMilli()})"
             },
             { issue ->
                 UpdateLinkedModule.Request(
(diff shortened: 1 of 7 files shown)


Marcono1234:
Thanks! Though #229 was only an idea, and there hasn't been much feedback on it yet.
So what do you others think about this change?

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
```diff
@@ -61,8 +63,10 @@ fun JiraIssue.getAttachments(remove: (JiraAttachment) -> Either<Throwable, Unit>
     attachments.map { it.toDomain(remove) }
 
 fun JiraVersion.toDomain(execute: (JiraVersion) -> Either<Throwable, Unit>) = Version(
+    name,
     isReleased,
     isArchived,
+    releaseDate.toYearInstant(),
```
NeunEinser, line 69: What's the difference to `toInstance()` (which we usually use here)

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -229,8 +230,8 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
         register(Modules.TransferVersions, TransferVersionsModule()) { issue ->
             AbstractTransferFieldModule.Request(
                 issue.key,
-                issue.getLinks(jiraClient, ::addAffectedVersionById, ::getVersionsGetField),
-                issue.versions.map { it.id }
+                issue.getLinks(jiraClient, ::addAffectedVersionById, ::getVersionsGetField.partially2 { Unit.right() }),
+                issue.getVersions { Unit.right() }
```
NeunEinser, line 234: What's with the hardcoded Unit.right()?

NeunEinser, line 234: What's with the hardcoded `Unit.right()`?
Is that unfinished code? It's supposed to return a mapped Version, IIRC

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
NeunEinser, line 69: What's the difference to `toInstance()` (which we usually use for this?

src/main/kotlin/io/github/mojira/arisa/domain/Version.kt
```diff
@@ -1,9 +1,12 @@
 package io.github.mojira.arisa.domain
 
 import arrow.core.Either
+import java.time.Instant
 
 data class Version(
+    val name: String,
```
NeunEinser, line 7: This needs to be kept as ID like before, otherwise the addVersionByID from jira/Calls.kt will not work

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
SPGoding: The `releaseDate` of version doesn't contain the part after day (it's `yyyy-MM-dd`). I was not sure if the regular one can parse that correctly so I created a new one. Will the regular one work if I replace the code with it?

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
SPGoding: It's passes as the `execute` function in `domain.Version` which is used to either add or delete the version from issue if I understand correctly. So I hardcoded it as `Unit.right()` here since we don't need either of that functions in this module.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
NeunEinser: Why wouldn't it?
It's still just an instance of Date.

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
NeunEinser: Right, we probably don't even *need* the remove by ID now. I'm cleaning this up a little

NeunEinser: Okay I understand this now, I am gonna work on #209 now and clean it up there

SPGoding: Can archived versions be removed? I've heard about that they can only be added but not removed, but I'm not really sure if that's true or not. If we _can_ remove them, #157 might need the remove by ID function.

SPGoding:
(not sure where to put this so I just leave it here since this was first mentioned under this PR)

> Right, we probably don't even need the remove by ID now.

How about removing future versions in the FutureVersion module?

NeunEinser:
It gets mapped to add here: 
https://github.com/mojira/arisa-kt/blob/master/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt#L201

Issue #142, opened 2020-04-22
Language module mis-fire


IonicEcko:
**Describe the bug**
The bug has very little information but none of it is non-English.

**Example**
https://bugs.mojang.com/browse/MCPE-71549

SPGoding:
Maybe Arisa shouldn't check tickets with too little words?

Marcono1234:
Another example: https://bugs.mojang.com/browse/MC-180009

IonicEcko:
Another example, this one longer than the last https://bugs.mojang.com/browse/MCPE-72386

NeunEinser:
Response:
```json
{
    "time": 1,
    "detectedLangs": [
        {
            "lang": "nl",
            "confidence": 0.936
        },
        {
            "lang": "pl",
            "confidence": 0.064
        }
    ],
    "timestamp": "2020-04-24T08:10:29.405"
}
```
It's 93.6% confident that this is Dutch, for whatever reason.

IonicEcko:
https://bugs.mojang.com/browse/MCPE-73476 another example

MikeUnverferth:
Does the language detection specifically ignore Minecraft-specific words, such as mob/block names? These at least are not likely to be in any language database and might throw off the detection.

urielsalis:
Supposedly it uses wikipedia as tokens and ignores those, im guessing Minecraft blocks are in wikipedia?

MikeUnverferth:
Often posters will repeat the summary in the description. Might help to check for this and exclude the summary if it is contained in the description (https://bugs.mojang.com/browse/MCPE-72386).

IonicEcko:
https://bugs.mojang.com/browse/MCPE-73975 This one doesn't seem to follow the 80% rule.

Marcono1234:
Maybe short words with latin chars like "I", "am", "to" and "do" are not unambiguous enough for the currently used service? The Polish chars are pretty unambiguous though I guess.
Therefore maybe it is rather confident that the text is Polish.

IonicEcko:
https://bugs.mojang.com/browse/MCPE-74498 another one

urielsalis:
Closing while preparing for release, all examples are instances of https://github.com/mojira/arisa-kt/issues/122

Marcono1234:
@urielsalis, no it appears not all of them are caused by that. @IonicEcko and @MikeUnverferth have provided examples where this occurred despite the complete report being in English:
[...]

Issue #122, opened 2020-04-18
The language module may close issues that include both a native Language and an English translation


NeunEinser:
**Describe the bug**
E.g. MC-179303.
The language API returns only French with a confidence of 100% for that ticket.
It might make sense to feed the API paragraphs individually and only close if none of them are English.

My haunch is that it saw non-English characters and thus ruled out English completely.

**Example**
[MC-179303](https://bugs.mojang.com/browse/MC-179303)

**Expected behavior**
It should not have closed that ticket.

PR #246, opened 2020-05-19
Double lengthThreshold of Language

SPGoding:
## Purpose
Try fixing #142... again...
## Approach
Statistics of known false positives on tickets that only has English texts (different from #122).
| Ticket     | Summary                                                 | Description                                                                                                                                     | Summary Length | Description Length | Total Length | Note                                                                     |
| ---------- | ------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- | -------------- | ------------------ | ------------ | ------------------------------------------------------------------------ |
| MC-180009  | Raining underground                                     | !image-2020-04-23-01-34-33-236.png \| thumbnail!                                                                                                | 19             | 47                 | 67           |                                                                          |
| MC-182790  | chickens despawn when their rider zombie dies           | chickens despawn when their rider zombie dies                                                                                                   | 45             | 45                 | 91           |                                                                          |
| MC-182947  | Ancient Debris does not generate in Basalt Deltas biome | Ancient Debris does not generate inside of basalt delta biomes                                                                                  | 55             | 62                 | 118          |                                                                          |
| MCPE-71549 | ?                                                       | Villagers can open iron doors                                                                                                                   | 1              | 29                 | 31           |                                                                          |
| MCPE-72386 | Baby Zombie Riding Adult Zombies                        | Baby Zombie Riding Adult Zombies\n\nZombie Jockey                                                                                               | 32             | 47                 | 80           |                                                                          |
| MCPE-73476 | Pillager doesn’t aim child villager                     | pillager doesn’t aim child villager.\n\n Reproduce:\n\n1.Summon pillager\n\n2.Summon child villager                                             | 35             | 93                 | 129          |                                                                          |
| MCPE-74498 | Zombie Villages generate about 25%-30%                  | Zombie Villages should spawn about 2%-5%.                                                                                                       | 38             | 41                 | 80           |                                                                          |
| MCPE-75920 | Armor Doesn't Render on Zombie Villager                 | Armor on Zombie Villager doesn't render but however the helmet renders on zombie zillager and also boots render but Not Chestplate or Leggings. | 39             | 143                | 183          | Broken English (no offense, still better than Bedrock's translations /s) |

I suppose doubling the threshold from `80` to `160` should be enough. Note this may create some false negatives (e.g. MCPE-79043 with length `115`, MCPE-79030 with length `155`), which is still better than false positives imo.

Do you have any ideas on how large the threshold shold be?

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -265,7 +265,7 @@
           "it": "I have detected that this ticket is in *Italian*."
         },
         "defaultMessage": "I would have closed this ticket as Invalid. This module is currently in the testing phase. If you find any issues, please [report them on GitHub|https://github.com/mojira/arisa-kt/issues].\nARISA_LANGUAGE_MODULE_TEST\n\n~{color:#888}-- I am a bot. This action was performed automagically!{color}~",
-        "lengthThreshold": 80
+        "lengthThreshold": 160
       },
       "hideImpostors": {},
       "resolveTrash": {


Marcono1234:
Note that the `lengthThreshold` applies to `"${(summary ?: "").trim()} ${(description ?: "").trim()}"`.
So even with this threshold change MCPE-75920 listed above (39 + 143 = 182) would still result in a false positive (182 > 160). However, that is probably acceptable because otherwise the false negative rate becomes too high (as pointed out in the description of this issue).

NeunEinser:
I looked at those individually.

* For MC-180009, we should just ignore formatting like that and remove attchment syntax. That would fall below the old threshold.
* For MC-182790, the API returns 0.5 confidence for bot German and English, when dots are added both after the summary and after the description.
* Additionally, we could make sure to check if summary and description are the same and not include the duplicate, which would make it fall below the old threshold as well.
* MC-18294 has always over 90% confidence of nl. If we check summary and and description for equal parts, it would be solved for that one in specific though.
* Not sure why MCPE-71549 is in there, it's already below the old threshold.
* MCPE-72386 would return all sorts of guesses (none over 80%) and thus fix the issue, if dots are inserted after the summary and at the end of the description
* MCPE-73476 Does return English with 56% if dots are inserted after description and summary and English with 92% if the summary is not sent (which is as the same first senstence of the description)
* MCPE-74498 Always returns nl with a high confidence, even with added punctuation after the description
* MCPE-75920 Always returns a confidence of 100% (?!) that it's nl.

So if we want to revoke this change I would propose:
* If the description contains the summary (ignoring case), the summary should not be sent.
* After summary and description a period should be added, if it isn't already present.
* nl should be whitelisted (This means, arisa will not detect dutch tickets anymore, but because of the many false positives related to that language, I think it is the best option)

SPGoding:
For MC-180009 and MCPE-71549, maybe they were created before the old threshold was added. Sorry for not filtering them out.

PR #247, opened 2020-05-19
Added issue as domain class, refcatored code

### LINKED ISSUE #209: Only map issues once per run

NeunEinser:
## Purpose
Resolves #209 

## Approach
Created an Issue domain class, to which the module executor maps after searching issues.
Since all relevant values are present in Issue now, Modules no longer have a Request data class, and are instead always called with an issue and lastRun now.

This currently breaks all tests.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/domain/Issue.kt
```diff
@@ -0,0 +1,47 @@
...
+    val createLink: (key: String, type: String) -> Either<Throwable, Unit>,
+    val addAffectedVersion: (id: String) -> Either<Throwable, Unit>,
+    val addEmptyComment: () -> Either<Throwable, Unit>,
+    val addModdedComment: () -> Either<Throwable, Unit>,
+    val addCrashDupeComment: (key: String) -> Either<Throwable, Unit>,
+    val addFutureVersionComment: () -> Either<Throwable, Unit>,
+    val addKeepPrivateComment: () -> Either<Throwable, Unit>,
+    val addNotEnglishComment: (lang: String) -> Either<Throwable, Unit>,
+    val addPiracyComment: () -> Either<Throwable, Unit>
+)
```
SPGoding, line 47: I suppose we can add default values to them? Honestly I don't know why I didn't do that in #248, that's another story anyways 🤦

NeunEinser: You can actually select how much context to show on review comments? HOW?!?!

Anyways, I don't know how much things overlap on other Domain classes, but it would probably be sane to have a meta-file for getX for this.

SPGoding: > You can actually select how much context to show on review comments? HOW?!?!

lol  
select

SPGoding:
Oops, what did I do 🤦

src/main/kotlin/io/github/mojira/arisa/domain/Issue.kt
NeunEinser: 👀 
New possibilities open up.
Mind is blown. 🤯

src/main/kotlin/io/github/mojira/arisa/domain/Issue.kt
```diff
@@ -0,0 +1,48 @@
...
+    val setPrivate: () -> Either<Throwable, Unit>,
+    val createLink: (key: String, type: String) -> Either<Throwable, Unit>,
+    val addAffectedVersion: (id: String) -> Either<Throwable, Unit>,
+    val addEmptyComment: () -> Either<Throwable, Unit>,
+    val addModdedComment: () -> Either<Throwable, Unit>,
+    val addCrashDupeComment: (key: String) -> Either<Throwable, Unit>,
+    val addFutureVersionComment: () -> Either<Throwable, Unit>,
+    val addKeepPrivateComment: () -> Either<Throwable, Unit>,
+    val addNotEnglishComment: (lang: String) -> Either<Throwable, Unit>,
+    val addPiracyComment: () -> Either<Throwable, Unit>
```
SPGoding, line 47: I think I'm going to change these stuff to four more general methods: `addComment(key)`, `addComment(key, filledText)`, `addRestrictedComment(key)`, and `addRestrictedComment(key, filledText)`. By doing so we will be able to test module-specific comment logic (like `addNotEnglishComment` and the upcoming crazy `addDuplicateComment`™) in each module's tests instead of somewhere else.

NeunEinser: I wanted to do that, but the helper messages thingie needs the project, so you can't just give the message when initializing the module. That's why I reverted it back to the old logic.

PR #248, opened 2020-05-19
Add functions with named parameters in tests

### LINKED ISSUE #242: Use named parameters for Jira Domain
Use named parameters for Jira Domain with defaults to avoid having to change all tests/modules when adding a new one

SPGoding:
## Purpose
Resolve #242.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis:
Awesome work!

src/test/kotlin/io/github/mojira/arisa/modules/AttachmentModuleTest.kt
```diff
@@ -35,7 +39,9 @@ class AttachmentModuleTest : StringSpec({
 
     "should return FailedModuleResponse when deleting fails" {
         val module = AttachmentModule(listOf(".test"))
-        val attachment = Attachment("testfile.test", NOW, { RuntimeException().left() }, { ByteArray(0) })
+        val attachment = getAttachment(
+            remove = { RuntimeException().left() }
+        )
```
NeunEinser, line 44: IIRC this can be `getAttachment() { RuntimeException().left() }`, because it's the last parameter and it's also a function

SPGoding:
My regular expression skills have been improved dramatically as I was massively replacing these codes 🤣

src/test/kotlin/io/github/mojira/arisa/modules/AttachmentModuleTest.kt
SPGoding: I suppose writing it as such can slightly improve readability and consistency with some other 
crazy domain classes like `Comment` which has two parameters (`restrict` and `update` ) of type function

urielsalis: I prefer making it explicit

PR #251, opened 2020-05-19
Support null releaseDate of versions

### LINKED ISSUE #250: Version's releaseDate can be null, causing FutureVersion failed
## Expected behavior
Only versions with a non-null `releaseDate` should be transfered.

SPGoding:
## Purpose
Fix #250.
## Approach
Only transfer versions that has a non-`null` `releaseDate`.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/TransferVersionsModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/TransferVersionsModule.kt
@@ -14,20 +14,28 @@ class TransferVersionsModule : AbstractTransferFieldModule<List<Version>, String
         field: List<Version>
     ) =
         parents.flatMap { (parentIssue, parentField) ->
-            val oldestVersionOnParent = getOldestVersion(parentField)
+            val oldestVersionWithKnownReleaseDateOnParent = getOldestVersionWithKnownReleaseDate(parentField)
             field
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/TransferVersionsModule.kt
```diff
@@ -14,20 +14,28 @@ class TransferVersionsModule : AbstractTransferFieldModule<List<Version>, String
...
+    /**
+     * Returns true only if:
+     * - The other version is `null`;
+     * - The other version has a `null` release date; OR
+     * - The current version has a non-`null` release date which is after the other version's release date.
+     */
     private infix fun Version.isReleasedAfter(other: Version?) =
-        other == null || this.releaseDate.isAfter(other.releaseDate)
+        other?.releaseDate == null ||
+                this.releaseDate?.isAfter(other.releaseDate) == true
```
NeunEinser, line 38:
```suggestion
                this.releaseDate?.isAfter(other.releaseDate)
```
please no `== true` 😂

SPGoding: This is intended, as that can be `null`

PR #252, opened 2020-05-19
Downgrade threshold to 80

SPGoding:
## Purpose
See [NE's comment](https://github.com/mojira/arisa-kt/pull/246#issuecomment-631014412)
## Approach
See [NE's comment](https://github.com/mojira/arisa-kt/pull/246#issuecomment-631014412)
## Future work
IDK
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/test/kotlin/io/github/mojira/arisa/modules/LanguageModuleTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/modules/LanguageModuleTest.kt
@@ -194,13 +194,14 @@ class LanguageModuleTest : StringSpec({
             "Description.",
             "not private",
             "private",
-            { it shouldBe "Summary. Description."; isApiExecuted = true; emptyMap<String, Double>().right() },
+            { it shouldBe "Summary. Description."; isApiExecuted = true; mapOf("en" to 1.0).right() },
             { Unit.right() },
             { Unit.right() }
         )
 
-        module(request)
+        val result = module(request)
 
+        result.shouldBeLeft(OperationNotNeededModuleResponse)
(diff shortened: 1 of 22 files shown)


src/test/kotlin/io/github/mojira/arisa/modules/LanguageModuleTest.kt
```diff
@@ -183,6 +183,114 @@ class LanguageModuleTest : StringSpec({
...
+            "Description.",
+            "not private",
+            "private",
+            { it shouldBe "Summary. Description."; emptyMap<String, Double>().right() },
+            { Unit.right() },
+            { Unit.right() }
+        )
+
+        module(request)
+    }
```
NeunEinser, line 201: You forgot the `result.shouldBeRight(ModuleResponse)` part in these

SPGoding: Intended actually, since I only want to test the text parsed to the API in these tests

urielsalis: I would test both in this case

NeunEinser: Then you need to make sure the function is actually called in some other way. E.g. by defining a variable at the top, setting it to the text in the function, and check the variable at the end.

SPGoding: That's a nice point!

PR #253, opened 2020-05-19
Add 'only' module setting

### LINKED ISSUE #219: Config to run a single module at the time for debugging
## The Problem
Just a nice to have, a way via config to only run a single module without needing to add a whitelist to all other modules

SPGoding:
## Purpose
Resolve #219

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -44,6 +44,10 @@ object Arisa : ConfigSpec() {
 
     object Modules : ConfigSpec() {
         open class ModuleConfigSpec : ConfigSpec() {
+            val only by optional<Boolean?>(
+                null,
```
urielsalis, line 48: can this be false instead of null?

PR #254, opened 2020-05-19
Resolve as awaiting response when removing future version

### LINKED ISSUE #228: Resolve reports with future version as "Awaiting Response" instead of replacing version
## The Problem
The FutureVersionModule currently removes future versions from the list of affected versions and instead adds (what it assumes to be) the latest released version. As discussed internally, this is not always ideal because the version which it adds as a replacement does not necessarily match the version in which the reporter is experiencing the issue in. See also #115.
This can then also cause issues when such a report is resolved as duplicate and the TransferVersionsModule then copies the incorrect versions. See also #229.

Side note: The module currently erroneously assumes that versions in the version list returned by Jira are sorted by release date, however at least for Bedrock edition this is not true. There "1.14.60 Hotfix" comes after "1.16.0.60 Beta" despite it being older. If this feature request is not implemented, the module should at least correctly sort by release date.

## Possible Solution
It was therefore suggested to resolve a report with future version as "Awaiting Response", keep the future version and let the reporter correct it (Arisa would then have to detect this) or let them clarify which version they are experiencing the problem in.
However, if the user added a valid version in addition to a future version, Arisa should only remove the future version and add a comment, but neither resolve the issue nor add an affected version itself.

urielsalis:
## Purpose
Fixes #228 

## Approach
Resolve issues as awaiting response after removing the future version. Includes a message telling the user why

For some reason this module doesn't work in MCTEST, probably on the versions check. We might want to ask them to mark one of the versions as released

#### Checklist
- [x] Included tests
- [x] Tested in https://bugs.mojang.com/browse/TRASH-25372

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -212,7 +212,7 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
                         issue.project.key, config[Modules.FutureVersion.message]
                     )
                 ),
-                ::resolveAs.partially1(issue).partially1("Invalid")
+                ::resolveAs.partially1(issue).partially1("Awaiting Response")
             )
         }
 
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -211,7 +211,8 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
                     messages.getMessageWithBotSignature(
                         issue.project.key, config[Modules.FutureVersion.message]
                     )
-                )
+                ),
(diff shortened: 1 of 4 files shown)


urielsalis:
image

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -211,7 +211,8 @@ class ModuleRegistry(jiraClient: JiraClient, private val config: Config, private
                     messages.getMessageWithBotSignature(
                         issue.project.key, config[Modules.FutureVersion.message]
                     )
-                )
+                ),
+                ::resolveAs.partially1(issue).partially1("Invalid")
```
SPGoding, line 215: I suppose you meant `Awaiting Response`

urielsalis: Its a good way to make sure we dont have the problem :P

fixed

Issue #231, opened 2020-05-17
[Idea] Verify helper-messages json during build


urielsalis:
## The Problem
We might deploy arisa when helper-messages doesn't contain a valid json
## Possible Solution
Verify it during the build?

SPGoding:
There's an [action](https://github.com/mojira/helper-messages/blob/master/.github/workflows/test-messages.yml) under helper-messages that validate the file. I'm not sure if it's sufficient though.

PR #255, opened 2020-05-19
Add test for helper messages and arisa.json

### LINKED ISSUE #148: Add a Github Action that validates arisa.json
## The Problem
It is possible to merge an invalid json, since `arisa.json` is not validated

## Possible Solution
A Github action that validates `arisa.json` should be added.
Similarily to mojira/helper-messages#34, a json schema should be added.

It might make sense to look into #100 first and change this issue accordingly.

urielsalis:
## Purpose
Test on build if arisa.json and helper-messages is valid

Fixes #148, #231

### DIFF
--- a/src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
@@ -1,0 +1,39 @@
+package io.github.mojira.arisa.infrastructure
+
+import com.uchuhimo.konf.Config
+import io.github.mojira.arisa.infrastructure.config.Arisa
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.booleans.shouldBeTrue
+import io.kotest.matchers.maps.shouldContainKey
+import java.io.File
+
+class IntegrationTest : StringSpec({
+    "should be able to read the JSON correctly" {
+        val config = Config { addSpec(Arisa) }
+            .from.map.flat(
+                mapOf(
+                    "arisa.credentials.username" to "test",
+                    "arisa.credentials.password" to "test",
+                    "arisa.credentials.dandelionToken" to "test"
+                )
+            )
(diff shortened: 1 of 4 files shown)


src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    "should contain required messages in helper-messages" {
+        val helperMessagesFile = File("helper-messages.json")
+        val helperMessages = helperMessagesFile.getHelperMessages()
+
+        with(helperMessages.messages) {
+            this shouldContainKey "pirated-minecraft"
+            this shouldContainKey "provide-affected-versions"
+            this shouldContainKey "incomplete"
+            this shouldContainKey "modified-game"
+            this shouldContainKey "panel-unmark-private-issue"
```
Marcono1234, line 36: If this should cover all used messages, would it make sense to somehow define these strings in one place only (possibly in `HelperMessages.kt` and then only permit using these)?
Otherwise I feel it would be easy to forget to update these when using more / different helper messages in the future.

src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+                    "arisa.credentials.username" to "test",
+                    "arisa.credentials.password" to "test",
+                    "arisa.credentials.dandelionToken" to "test"
+                )
+            )
+            .from.json.watchFile("arisa.json")
+            .from.env()
+            .from.systemProperties()
+
+        config.containsRequired().shouldBeTrue()
```
NeunEinser, line 24: I think it would make sense to require almost all then.
Currently most things are still optional, but the module doesn't really do anything without them

urielsalis: We should make things that the module breaks without required

PR #261, opened 2020-05-20
Comment resolveAsInvalid out in LanguageModule

SPGoding:
## Purpose
Make LanguageModule not actually resolve tickets.
## Approach
Comment them out.
#### Checklist
- [x] Excluded tests
- [ ] Tested in MCTEST-xxx


SPGoding:
> When was this uncommented?

It was always uncommented. Before #247 we passed a `{ Unit.right }` function in the `Request` for `LanguageModule` in `ModuleRegistry`, so Arisa didn't actually resolve the ticket. After that refactor every module uses the same `resolveAsInvalid` function, so we need to comment that out ;/

NeunEinser:
No, I specifically commented it

PR #264, opened 2020-05-21
Group API requests

### LINKED ISSUE #192: Arisa adds a single changelog entry per action
## The Problem
Arisa adds a single changelog entry for every single action that it takes. That can get quite spammy at times.

This affects every module that does more than one action as far as I can tell.

## Example ticket
https://bugs.mojang.com/browse/MC-159283

## Possible Solution
Group API requests. This is possible because we as moderators can do that via the website. This is probably not easy to implement.

SPGoding:
## Purpose
Fix #192.
## Approach
Added a new cache object named `IssueUpdateContextCache`, which stores a map from issue keys to the necessary information (stored in `IssueUpdateContext`) we need to make an update and a transition. Functions in `infrastructure.jira.Functions` now accept the `IssueUpdateContext` class instead of the `JiraIssue` class.

At the end of each run we will apply the updates and transitions of every `IssueUpdateContext` stored in the cache, and then clear the cache.

Changes that use the same update method and are run in the same run will be grouped after merging this PR:
| Field / Operation | Update method | Note |
| - | - | - |
| CHK | transition |
| Confirmation Status | transition |
| Linked | transition |
| security | transition |
| resolution | transition |
| reopen | transition |
| resolveAs | transition |
| description | update |
| Affected Version/s | update | Technically this can also be done by transition, however the library doesn't provide `fieldAdd` for transitions. |
| createComment | N/A | Technically this can also be done by transition, however the library doesn't provide `fieldAdd` for transitions. |
| addRestrictedComment | N/A | Technically this can also be done by transition, however the library doesn't provide `fieldAdd` for transitions. |
| deleteAttachment | N/A |
| createLink | N/A |
| deleteLink | N/A |
| updateCommentBody | N/A |
| restrictCommentToGroup | N/A |
| getGroups | N/A |
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContextCache.kt
```diff
@@ -0,0 +1,5 @@
+package io.github.mojira.arisa.domain
+
+import io.github.mojira.arisa.infrastructure.Cache
+
+object IssueUpdateContextCache : Cache<IssueUpdateContext>()
```
NeunEinser, line 5: Why is this needed?

src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContext.kt
```diff
@@ -0,0 +1,11 @@
+package io.github.mojira.arisa.domain
+
+import net.rcarz.jiraclient.Issue
+
+data class IssueUpdateContext(
+    val update: Issue.FluentUpdate,
+    val transition: Issue.FluentTransition,
+    var hasUpdates: Boolean = false,
+    var hasTransitions: Boolean = false,
+    var transitionName: String = "Update Issue"
```
NeunEinser, line 10: This currently does only support one transition.
It is conceivable that we will need both an update and another transition at some point though.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -18,48 +20,93 @@ import java.net.URI
...
+}
+
+fun updateSecurity(context: IssueUpdateContext, levelId: String) {
+    context.hasTransitions = true
+    context.transition.field(Field.SECURITY, Field.valueById(levelId))
+}
+
+fun removeAffectedVersion(context: IssueUpdateContext, version: Version) {
+    context.hasUpdates = true
+    context.update.fieldRemove("versions", version)
```
NeunEinser, line 81: To avoid having both an update transition and a regular update, this could be moved to the update transition instead.
This is true for everything in a regular update, except of the description.

If you'd like to avoid the change being denoted as "Opened", "Reopened" or "Resolved" in emails and logs and and avoid having the status field show up in the changelog, you could do a bit more of an advanced approach and have a registry where fields are regisered with booleans whether it's possible to change them in an update and an update transiton, and prefer an update, when everything is possible as a simple update.

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -55,6 +57,9 @@ class ModuleExecutor(
                     )
                 }
 
+                IssueUpdateContextCache.forEach(::applyUpdatesAndTransitions)
+                IssueUpdateContextCache.clear()
```
NeunEinser, line 61: This is currently in the executeModules method, which means that in case one function or one module does something wrong, the entire run will fail and last-run will remain at the old time.

This should move into each module's execute function and instead return a failed module response for that module, so that we only save issue keys where modules actually failed and don't have to run through all issues if something bad happened.

src/main/kotlin/io/github/mojira/arisa/infrastructure/Cache.kt
```diff
@@ -0,0 +1,17 @@
+package io.github.mojira.arisa.infrastructure
+
+@Suppress("MemberNameEqualsClassName")
```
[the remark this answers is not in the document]
SPGoding: Rip my vocabulary.

src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContextCache.kt
SPGoding: Because I don't feel like I want to pass the cache instance to every mapper function, so I just create it as a singleton.

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
SPGoding: Then that will not group these API requests if I understand you correctly

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
SPGoding: :cry: Sadly the Jira library we use doesn't provide a `fieldAdd` and `fieldRemove` method for transition, as I noted in the PR summary

src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContextCache.kt
urielsalis: I think its better to pass it, that way we can swap implementations in a single place instead of multiple

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
NeunEinser: That's exactly what happens when one module fails.
If you can adapt that code to collect failed tickets here, that would be fine as well.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
NeunEinser: This lib also doesn't support deleting attachments, we had to write our own API call for that.
Does the official one not provide the rest client?

Also I think fieldAdd and fieldRemove could still work by setting the field to all previous entries plus (or minus for remove) one entry. Would be kinda annoying to handle for multiple adds/removes, though.

src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContext.kt
NeunEinser: Wdym?
The regular updates, yes.
But we have an update transition here. That's a custom transition on our tracker and has nothing to do with the normal update.

The normal update is just a put on the issue, while the update transition is something custom.

NeunEinser: Oh do you mean that you can add fields tro any transition?

You can only add fields to a transition that are available on the screen, otherwise you get a 400 (or something).

For resolve the screen only contains what's visible here:
image

The update transition is the only one that supports these fields.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
SPGoding: No, the official one doesn't even expose the rest client (or at least I can't find it anywhere)

src/main/kotlin/io/github/mojira/arisa/domain/IssueUpdateContext.kt
SPGoding: Good to know!

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
SPGoding: > If you'd like to avoid the change being denoted as "Opened", "Reopened" or "Resolved" in emails and logs and and avoid having the status field show up in the changelog

iirc updates don't trigger any email, so we don't even to worry about that, right? And I personally don't really care about that status change shows in the changelog since they are anyways in the same item.

SPGoding: > Also I think fieldAdd and fieldRemove could still work by setting the field to all previous entries plus (or minus for remove) one entry.

I gave up lol

NeunEinser: > 
> 
> > If you'd like to avoid the change being denoted as "Opened", "Reopened" or "Resolved" in emails and logs and and avoid having the status field show up in the changelog
> 
> iirc updates don't trigger any email, so we don't even need to worry about that, right? And I personally don't really care about that status change shows in the changelog since they are anyways in the same item.

Yeah, both methods have advantages and disadvantages, that's why I don't have really have a preference.

NeunEinser: Actually, if we want to add versions in update this, it would probably be the easiest to just use our own fork from the (current) lib and implement these things there, instead of constantly working around them

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -18,142 +23,178 @@ import java.net.URI
...
+    val resolutionJson = JSONObject()
+    resolutionJson["name"] = resolution
 
-fun reopenIssue(issue: Issue) = runBlocking {
-    Either.catch {
-        issue.transition().execute("Reopen Issue")
-    }
+    context.resolve.field(Field.RESOLUTION, resolutionJson)
+    context.transitionName = "Resolve Issue"
 }
```
NeunEinser, line 73: You are executing everything from all modules after a run, right?
What happens if there are two conflicting resolutions (e.g. Duplicate from CrashModule and Invalid from PiracyModule)?

SPGoding: The one that is changed later will take effects, I believe

NeunEinser: Looked at the lib code, seems to be a regular map, so the last put should indeed be the one that takes effect.

We might want to prefer Dupe over everything else, as that's usually the most helpful, but since it probably still does the linking and post all the comments, it should really be fine.
We can always change it

Marcono1234:
The build failure after the merge seems to be caused by JitPack: https://github.com/jitpack/jitpack.io/issues/4226

PR #265, opened 2020-05-21
Feature/duplicate message

### LINKED ISSUE #116: Add "duplicates" comment when it does not exist
There exists a helper message template for when a report is a duplicate, see https://mojira.github.io/helper-messages/:
> ...
We're actually already tracking this issue in **MC-123**, so I've resolved and linked this ticket as a duplicate.
...

However, adding that manually is tedious.
Maybe it would be good if Arisa automatically added it, either always or if the reporter is new (e.g. only has a few reports) so they learn about the search feature. Arisa would have to choose the correct template based on the project then because some links such as the "Issue Guidelines" link are project specific. And it should not add a comment if one already exists. This might be a little bit tricky because sometimes non-staff members add such a comment.

If this functionality was added, all staff members would have to be informed to prevent them adding a duplicate comment by accident when Arisa already added one.

Maybe this effort is not worth it though, what do you think?

SPGoding:
## Purpose
Resolve #116.
## Approach
After merging this PR, we will be able to make Arisa send different duplicate messages for tickets with normal parents, private parents, parents with specific resolutions, and specific parents. Arisa will not send the message only if all of the parent keys have been mentioned publicly in former comments.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -63,6 +63,7 @@
         "resolutions": [
           "Duplicate"
         ],
+        "commentDelay": 5,
         "message": "duplicate",
         "ticketMessages": {
           "MC-297": "duplicate-of-mc-297",
--- a/arisa.json
+++ b/arisa.json
@@ -59,6 +59,17 @@
           "url"
         ]
       },
+      "duplicateMessage": {
+        "resolutions": [
+          "Duplicate"
+        ],
+        "message": "duplicate",
+        "privateMessage": "duplicate-private",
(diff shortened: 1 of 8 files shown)


Marcono1234:
Would it be possible to add some delay (as mentioned in https://github.com/mojira/arisa-kt/issues/116#issuecomment-631159748) to prevent it from commenting when a moderator wants to comment a custom message?

PR #268, opened 2020-05-21
Not reopen AR when people other than the reporter comment

### LINKED ISSUE #170: Don't reopen old "Awaiting Response" reports if person other than reporter updates it
## The Problem
One issue #38 tries to solve is that reports which have been resolved as "Awaiting Response" and haven't been updated for a while get reopened by other users which add unrelated comments or attachments. And chances are high that this action is missed and the report therefore incorrectly remains open.

This issue here proposes a possible solution to this problem while not going as far as #38 and requesting resolution changes of the report.

## Suggestion
Only reopen "Awaiting Response" reports if the following applies:
- The report was resolved less than _X_ months ago (can be queried through `resolutiondate` field in API response?, assuming it updates when re-resolving)
- **OR:** The person who performed the update (adding comment or attachment or updating report)  is the reporter

If this does not apply, Arisa should not reopen the report, but instead comment something similar to:
> This report has been resolved a rather long time ago and will not be reopened automatically. If a moderator sees your comment, they will reopen and update the report. However, if you think your update to this report has been overlooked or you want to make sure that this report is reopened, you can contact the Mojira staff on /r/Mojira or the Discord server.

Note: Updates by staff should already be filtered out before and not trigger reopening or the addition of the comment, see also #117.

SPGoding:
## Purpose
Resolve #170.
Depends on https://github.com/mojira/helper-messages/pull/60

## Approach
Added a new `softARPeriod`. The ticket can be reopened by comments posted by people other than the reporter within `softARPeriod` days after it was resolved. After the time has passed, only the reporter can reopen the ticket.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -156,6 +156,12 @@ object Arisa : ConfigSpec() {
                 null,
                 description = "A tag used to indicate that Arisa should keep the ticket Awaiting Response"
             )
+            val softARPeriod by optional<Long?>(
+                null,
+                description = "The ticket can only be reopened by comments posted by people other than the reporter " +
```
Marcono1234, line 161:
```suggestion
                description = "The ticket can also be reopened by comments posted by people other than the reporter " +
```
"Only" would mean that the reporter is not able to reopen it that way.

arisa.json
```diff
@@ -135,7 +135,9 @@
           "staff",
           "global-moderators"
         ],
-        "keepARTag": "MEQS_KEEP_AR"
+        "softARPeriod": 365,
```
urielsalis, line 138: can you add the time unit in the name?

src/test/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModuleTest.kt
```diff
@@ -500,12 +652,67 @@ class ReopenAwaitingModuleTest : StringSpec({
...
         )
 
         val result = MODULE(issue, TEN_SECONDS_AGO)
 
         result.shouldBeRight(ModuleResponse)
+        hasReopened shouldBe false
+        hasCommented shouldBe true
     }
 
     "should return FailedModuleResponse with all exceptions when reopening fails" {
```
NeunEinser, line 718: I think the Failed tests are missing for when it fails to comment.

arisa.json
NeunEinser:
```suggestion
        "softARDays": 365,
```
For example

src/test/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModuleTest.kt
SPGoding: We won't need them after #264 anyways

NeunEinser: Yeah this was posted slightly before that pr, or at least before I became aware of it

SPGoding:
Yes :p

Issue #227, opened 2020-05-11
Set issues containing an email address to private 


NeunEinser:
## The Problem
It happens frequently that account issues are submitted containing an email adress but the reporter doesn't set it to private. Arisa could set those automatically to ptivate.
It shouldn't leave a keep private comment though, so it still pops up in filters and human mods can take a look at it and set it to public again, when aplicable.

I am also thinking of the Language module which currently sends all data to a third party server, if they are not private. When this is implemented, emails would also not be sent to them

## Possible Implementation
A regex should check the summary, enviornment and issue text and search for any valid email adresses.
The module should only work on issues that have been created since the last run (in case a moderator purposefully set it to public, Arisa should not set it to private again).

If applicable it should then set the security level to private, similarily as the keep private module does.

Marcono1234:
> I am also thinking of the Language module which currently sends all data to a third party server, if they are not private. When this is implemented, emails would also not be sent to them

Would this really work? I am not familiar enough with Arisa, but would the language module see when this module set the report to private in the same run (assuming order of module execution is already correct)?

PR #270, opened 2020-05-22
Add PrivacyModule

### LINKED ISSUE #53: Prevent users from publishing their session id
Even though the launcher and the client were changed to not include the session id anymore, there are still people playing with old launchers old versions of the game (or the new launcher does not censor it?).

It would be good if Arisa could prevent people from

putting the session id in the description
writing the session id in a comment
attaching a log file which contains the session id
The log line would have the prefix (Session ID is and would be (Session ID is <censored>) if it was censored by the old Java launcher.

Moved from https://github.com/mojira/arisa-js/issues/12

SPGoding:
## Purpose
Resolve #53 and #227
## Approach
Add a new PrivacyModule which checks newly created issue's summary, environment, and description; newly created public comments; and newly uploaded text attachments.

**Require mojira/helper-messages#80**
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -0,0 +1,54 @@
...
+                .map { { it.restrict(it.body) } }
+                .toList()
+
+            assertEither(
+                assertTrue(stringMatchesPatterns),
+                assertNotEmpty(restrictCommentFunctions)
+            ).bind()
+
+            if (stringMatchesPatterns) {
+                setPrivate().toFailedModuleEither().bind()
```
urielsalis, line 46: Can it leave a comment if its a ticket or edit the comment to leave a messagei f its a comment?

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
+        Either.fx {
+            assertNull(securityLevel).bind()
+
+            var string = ""
+
+            if (created.isAfter(lastRun)) {
+                string += "$summary $environment $description "
+            }
```
NeunEinser, line 22: It might make sense to also scan changelog entries created after last run, and also resolve if fromString does not succeed, but toString does.

urielsalis:
Is this still being worked on?

SPGoding:
Yes, I will do this after #290

SPGoding:
I have read the CLA Document and I hereby sign the CLA

urielsalis:
Does the pattern you are using for emails work on emails with . and tlds with less or more characters than 3?

SPGoding:
Yes, it also recognizes something like `!@(^#(*.#%&(` which should be fine I suppose.

PR #273, opened 2020-05-22
Fix duplicate message commentDelay

SPGoding:
## Purpose
Fix `commentDelay` not working
## Approach
Whatever 🤔
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


SPGoding:
It's really late in my timezone now, I will debug this strange stuff tommorrow

I wish Arisa wouldn't cause too many _duplicated_ duplicate comments ;D

Issue #259, opened 2020-05-20
Create command parser


urielsalis:
## The Problem
Create a command parser, using something like https://github.com/Mojang/brigadier

This will allow us to create commands to do things like adding archived versions (see #157) in a easier way to adding a command

PR #279, opened 2020-05-23
Add Add archived version command

### LINKED ISSUE #277: Add command for adding affected versions
## The Problem
Split from #157

## Possible Solution

Add a command, triggered by a volunteer adding a staff comment with "ARISA_ADD_VERSION 12w34a"

urielsalis:
## Purpose
Fix #277,  #259, #280
## Approach
Implement simple command parser and a command
## Future work
- If we add more commands it might be worth creating a registry
- Fixed command, its a different issue
- It works as an example on how a test for module executor could look like :)
#### Checklist
- [x] Included tests
- [x] Tested in MC-185329, MCTEST-88

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
@@ -1,0 +1,21 @@
+package io.github.mojira.arisa.modules.commands
+
+import arrow.core.Either
+import arrow.core.extensions.fx
+import arrow.core.left
+import io.github.mojira.arisa.domain.Issue
+import io.github.mojira.arisa.modules.ModuleError
+import io.github.mojira.arisa.modules.ModuleResponse
+import io.github.mojira.arisa.modules.OperationNotNeededModuleResponse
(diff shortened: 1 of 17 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.github.mojira.arisa.domain.Issue
+import io.github.mojira.arisa.modules.ModuleError
+import io.github.mojira.arisa.modules.ModuleResponse
+import io.github.mojira.arisa.modules.OperationNotNeededModuleResponse
+import io.github.mojira.arisa.modules.assertTrue
+
+class FixedCommand : Command {
+    override fun invoke(issue: Issue, vararg arguments: String): Either<ModuleError, ModuleResponse> = Either.fx {
+        assertTrue(arguments.size > 1).bind()
+        val version = arguments[1]
```
Marcono1234, line 15:
```suggestion
        val version = arguments.asList().subList(1, arguments.size).joinToString(" ")
```
Probably same as for AddVersionCommand.kt to handle version names with spaces.

PR #290, opened 2020-05-27
Make context lazy

SPGoding:
## Purpose
Fix taking up too much RAM
## Approach
Makes context lazy
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis:
58 files, nice

SPGoding:
> 58 files, nice 

Lol, this PR contains everything #289 has

src/main/kotlin/io/github/mojira/arisa/domain/Link.kt
```diff
@@ -1,10 +1,8 @@
 package io.github.mojira.arisa.domain
 
-import arrow.core.Either
-
 data class Link(
     val type: String,
     val outwards: Boolean,
     val issue: LinkedIssue,
-    val remove: () -> Either<Throwable, Unit>
```
urielsalis, line 9: We are saying this cant fail anymore? I dont think thats correct?
Same for all other functions

urielsalis:
I recommend you look into promises for this kind of stuff. It allows you to mark something as running later but still potentially failing and handle it

src/main/kotlin/io/github/mojira/arisa/domain/Link.kt
SPGoding: All API requests are sent at the end of each run; what the `remove` function here does is to add the remove operation to the update context's `otherOperations` list, which will never fail

SPGoding:
> I recommend you look into promises for this kind of stuff. It allows you to mark something as running later but still potentially failing and handle it

I believe promises are for asynchronous purposes, which isn't the case here. Every ticket that is mapped into a domain has its own `IssueUpdateContext`, however most of the tickets won't be updated and therefore the context is unnecessarily taking up the RAM. This PR changes the type of context to `Lazy<IssueUpdateContext>` so the context object will only be created when needed

src/main/kotlin/io/github/mojira/arisa/domain/Link.kt
urielsalis: what I mean is that at the end of the run it might fail, and if it does we might not want to have done other tasks (like commenting if it failed resolving, as it will just start spamming comments). Thats where a promise might help. Its a operation that will be executed async at the end of the run but might still fail so we might want to execute things if it fails, things if it doesnt

SPGoding: Ah I think I get you now. Then what should we do when it fails so that we can prevent it from spamming comments in your example?

urielsalis: It requires a change on how we are doing things now, but:

1- We make it do all actions in the same module, so multiple modules have multiple caches
2- The functions return a promise, and we combine the multiple actions as multiple promises. Then we do a .onError {} after they executed to do one action(if the module requires), flatMap to do something if everything else is succesful(where we might do more jira request that dont go through the cache)

Honestly 1 is simpler to implement, but it would cause multiple changes, can we calculate how many changes in both options we save in common cases?

SPGoding: The first option can already resolve the multiple changes problem caused by the `TransferVersionModule` mentioned in #192.

The second option can combine changes to fields like `Linked` and `Confirmation Status` furthermore, which I honestly don't think worth the work ;P

And if we decide to take the first option, how do you feel about moving the applying-changes-and-clearing-cache part to the `executeModule` function?

urielsalis: it would be better, yes :D

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -65,10 +68,10 @@ class ModuleExecutor(
         }
     }
 
-    @Suppress("LongParameterList")
+    @Suppress("LongMethod", "LongParameterList")
```
urielsalis, line 71: We should probably split this function instead 😓

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -52,15 +52,15 @@ class CrashModule(
 
             if (key == null) {
                 if (anyModded) {
-                    addComment(CommentOptions(moddedMessage)).toFailedModuleEither().bind()
-                    resolveAsInvalid().toFailedModuleEither().bind()
+                    addComment(CommentOptions(moddedMessage))
```
urielsalis, line 55: add comment should be always after resolving, and it should only happen if the module didnt fail to do the resolve

src/main/kotlin/io/github/mojira/arisa/modules/EmptyModule.kt
```diff
@@ -37,8 +37,8 @@ class EmptyModule(
                 assertNotEqual(environment, ENV_DEFAULT).bind()
             }
             assertEmpty(attachments).bind()
-            addComment(CommentOptions(message)).toFailedModuleEither().bind()
-            resolveAsIncomplete().toFailedModuleEither().bind()
+            addComment(CommentOptions(message))
```
urielsalis, line 40: add comment should be always after resolving, and it should only happen if the module didnt fail to do the resolve

src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
```diff
@@ -21,10 +21,10 @@ class FutureVersionModule(
             val latestVersion = project.versions.lastOrNull(::isFutureVersion.complement())
             assertNotNull(latestVersion).bind()
 
-            latestVersion!!.add().toFailedModuleEither().bind()
-            tryRunAll(removeFutureVersions).bind()
-            addComment(CommentOptions(message)).toFailedModuleEither().bind()
-            resolveAsAwaitingResponse().toFailedModuleEither().bind()
+            latestVersion!!.add()
+            removeFutureVersions.forEach(::run)
+            addComment(CommentOptions(message))
```
urielsalis, line 26: add comment should be always after resolving, and it should only happen if the module didnt fail to do the resolve

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -23,9 +23,9 @@ class KeepPrivateModule(
             val markedTime = comments.first(::isKeepPrivateTag).created
             val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
             if (changedTime != null && changedTime.isAfter(markedTime)) {
-                addComment(CommentOptions(message)).toFailedModuleEither().bind()
+                addComment(CommentOptions(message))
```
urielsalis, line 26: add comment should be always after resolving, and it should only happen if the module didnt fail to do the resolve

src/main/kotlin/io/github/mojira/arisa/modules/PiracyModule.kt
```diff
@@ -18,8 +18,8 @@ class PiracyModule(
                 piracySignatures,
                 "$description $environment $summary"
             ).bind()
-            addComment(CommentOptions(message)).toFailedModuleEither().bind()
-            resolveAsInvalid().toFailedModuleEither().bind()
+            addComment(CommentOptions(message))
```
urielsalis, line 21: add comment should be always after resolving, and it should only happen if the module didnt fail to do the resolve

src/main/kotlin/io/github/mojira/arisa/modules/TransferLinksModule.kt
```diff
@@ -28,8 +27,8 @@ class TransferLinksModule : AbstractTransferFieldModule() {
     private fun parentDoesNotHaveLink(parentLinks: List<Link>, other: Link) =
         parentLinks.none {
             it.type == other.type &&
-                it.outwards == other.outwards &&
-                it.issue.key == other.issue.key
+                    it.outwards == other.outwards &&
```
urielsalis, line 30: You have continuation space set to 8?

src/main/kotlin/io/github/mojira/arisa/modules/TransferVersionsModule.kt
```diff
@@ -24,7 +22,7 @@ class TransferVersionsModule : AbstractTransferFieldModule() {
                 .filter { it isReleasedAfter oldestVersionWithKnownReleaseDateOnParent }
                 .map { it.id }
                 .filter { it !in parentVersionIds }
-                .map(parent.addAffectedVersion::partially1)
+                .map { { parent.addAffectedVersion(it) } }
```
urielsalis, line 25: 👏

src/main/kotlin/io/github/mojira/arisa/modules/TransferLinksModule.kt
SPGoding: Yes, which I really don't like lol. Changed to 4.

PR #293, opened 2020-05-30
Add test for Module Registry

urielsalis:
## Purpose
Tests the module registry to make sure we dont forget a module
## Approach
Check that we have a registered module per config
Check that we have a registered module per module class
Check that the configs are used by only 1 module to prevent us forgetting from using the correct config

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -72,7 +72,7 @@ class ModuleRegistry(private val config: Config) {
     ) = { issue: Issue, lastRun: Instant ->
         config::class.simpleName!! to
                 ({ lastRun pipe (issue pipe2 module::invoke) } pipe ::tryExecuteModule)
-    } pipe (getJql pipe2 (config pipe3 (config::class.java.simpleName pipe4 ModuleRegistry::Entry))) pipe modules::add
+    } pipe (getJql pipe2 (config pipe3 (config::class.simpleName!! pipe4 ModuleRegistry::Entry))) pipe modules::add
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -70,7 +72,7 @@ class ModuleRegistry(private val config: Config) {
     ) = { issue: Issue, lastRun: Instant ->
         config::class.simpleName!! to
                 ({ lastRun pipe (issue pipe2 module::invoke) } pipe ::tryExecuteModule)
-    } pipe (getJql pipe2 (config pipe3 ModuleRegistry::Entry)) pipe modules::add
+    } pipe (getJql pipe2 (config pipe3 (config::class.java.simpleName pipe4 ModuleRegistry::Entry))) pipe modules::add
```
NeunEinser, line 75: Is `.java` needed here? Previously, it was without that.

src/test/kotlin/io/github/mojira/arisa/ModuleRegistryTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        println("Not unique modules " + modules.filter { !uniqueModules.contains(it) })
+        uniqueModules shouldContainExactlyInAnyOrder modules
+    }
+
+    "should get only get a module with only" {
+        val modules = ModuleRegistry(getConfig(true)).getModules()
+        modules.size shouldBe 1
+    }
+
+    "should register a module for each file ending in Module" {
```
NeunEinser, line 32: [Minor] Maybe change the description to `should register a module for each non-abstract module`, to also cover Abstract...?
(Also not sure if you can use reflections to look for abstract, maybe we don't need to rely on the file name)

PR #315, opened 2020-06-20
Add command for fixing tickets with a specific fixed version

### LINKED ISSUE #278: Add command for adding fixed versions
## The Problem
Split from #157

## Possible Solution

Add a command, triggered by a volunteer adding a staff comment with "ARISA_FIXED 12w34a"

Discussion: should this be allowed for helpers?

urielsalis:
## Purpose
Fix #278

## Approach
Transition to fixed adding the version. Unfortunately update(to add multiple) does not work in MC due to lack of permissions, so we need to start with a open issue

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-95


NeunEinser:
Since the CommandModule seems to allow both helpers and moderators, this would allow helpers to resolve a ticket as fixed.
Maybe the command interface should provide a way to specify restriction levels?

PR #326, opened 2020-06-20
Fix email detection

### LINKED ISSUE #325: Improve email pattern in PrivacyModule
## The Bug
`{{@e[team=...]}}` matches the Email pattern (`\S+@\S+\.\S{2,}`)
## Example Ticket
https://bugs.mojang.com/browse/MC-190703
## Expected behavior
https://emailregex.com/

urielsalis:
## Purpose
Fix #325
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis:
wordpatterns runs word by word, fulltextpatterns runs on the whole text

SPGoding:
I still can't see why you need to do that, as the email regex doesn't allow any spaces in it afaik so it works exactly the same to run it word by word or on the whole text directly

PR #331, opened 2020-06-21
Prevent from spamming comments

### LINKED ISSUE #190: Prevent Arisa from spamming a ticket
## The Problem
In order to avoid things like #189 from happening again, there should probably be a check somewhere that prevents Arisa from doing the same action on the same ticket multiple times in a row.

## Possible Solution
Keep track of when Arisa does an action on a ticket in a database (or similar), and if there is a recent entry for an action that is about to be done, don't do it (and log a warning, because it indicates a malfunctioning module).

Alternatively Arisa could ignore ticket updates where the last changelog entry is from itself, but that could cause other issues, for example since multiple modules may apply to the same ticket.

SPGoding:
## Purpose
Resolve #190.
## Approach
Keep a cache which stores the comments that are posted under each ticket in the last run.
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
@@ -5,6 +5,8 @@ package io.github.mojira.arisa.infrastructure.jira
 import arrow.core.Either
 import arrow.syntax.function.partially1
 import io.github.mojira.arisa.domain.IssueUpdateContext
+import io.github.mojira.arisa.infrastructure.Cache
+import io.github.mojira.arisa.log
 import io.github.mojira.arisa.modules.FailedModuleResponse
 import io.github.mojira.arisa.modules.ModuleResponse
 import io.github.mojira.arisa.modules.tryRunAll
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -141,11 +143,26 @@ fun deleteAttachment(context: Lazy<IssueUpdateContext>, attachment: Attachment)
...
+) {
     context.value.otherOperations.add {
         runBlocking {
             Either.catch {
-                context.value.jiraIssue.addComment(comment)
+                val key = context.value.jiraIssue.key
+                val oldPostedComments = oldPostedCommentCache.get(key)
+                val newPostedComments = newPostedCommentCache.getOrAdd(key, mutableSetOf())
+
+                if (oldPostedComments != null && oldPostedComments.contains(comment)) {
```
urielsalis, line 159: Lets imagine a comment is posted, then next run it isnt. What would happen on the third run? It would be posted or not?

SPGoding: Let's assume the comment has been posted in the first run.

- If it is never tried to be posted in the second run, it will be posted in the third run.
- If it tried to be posted but actually not posted because of the preventing-spam mechanism, it will still _not_ be posted in the third run.

PR #350, opened 2020-07-01
Add DiscordAppender for logback

### LINKED ISSUE #348: Forward log to a Discord channel
## The Problem
Contributors besides the chosen ones can't enjoy the fancy ~~crash~~ logs 😭

SPGoding:
## Purpose
image

Resolve #348, by the most desired way.

> If that's possible, it would be the best solution, every time something is logged, push that line / these lines in text form to a webhook, so that we essentially have the entire log on Discord always.
>
> _Originally posted by [~NeunEinser] in https://github.com/mojira/arisa-kt/issues/348#issuecomment-652571629_

## Approach
By adding a new [DiscordAppender](https://github.com/napstr/logback-discord-appender) in `logback.xml`.

## Future work
- [x] [Create two new channel](https://support.discord.com/hc/en-us/sections/201110547-Channel-Setup) in the Mojira Discord server.
- [x] [Create two webhooks](https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks) for the channels. The name and the avatar of the webhooks don't matter, as they will be overridden by the payload.
- [ ] Add the link of the webhooks to `arisa.yml`'s `arisa.credentials.discordLogWebhook` and `arisa.credentials.discordErrorLogWebhook`.
```diff
 arisa:
   logOperationNotNeeded: true
   credentials:
     username: arisabot
     password: xxx
     dandelionToken: xxx
+    discordLogWebhook: https://discordapp.com/api/webhooks/123/xxx
+    discordErrorLogWebhook: https://discordapp.com/api/webhooks/456/xxx
```
4. Merge this PR.
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -20,7 +20,8 @@ class ModuleExecutor(
     private val queryCache: Cache<List<Issue>>,
     private val issueUpdateContextCache: IssueUpdateContextCache,
     private val searchIssues:
-        (Cache<MutableSet<String>>, Cache<MutableSet<String>>, String, Int, () -> Unit) -> List<Issue>
+        (Cache<MutableSet<String>>, Cache<MutableSet<String>>, String, Int, () -> Unit) -> List<Issue>,
+    private val doLog: Boolean = true
 ) {
     private var postedCommentCache = Cache<MutableSet<String>>()
 
@@ -63,7 +64,7 @@ class ModuleExecutor(
             } while (missingResultsPage)
             return ExecutionResults(true, failedTickets)
         } catch (ex: Throwable) {
-            log.error("Failed to execute modules", ex)
+            if (doLog) log.error("Failed to execute modules", ex)
             return ExecutionResults(false, failedTickets)
         } finally {
             postedCommentCache = newPostedCommentCache
@@ -90,18 +91,18 @@ class ModuleExecutor(
                 response.second.fold({
                     when (it) {
                         is OperationNotNeededModuleResponse -> if (config[Arisa.logOperationNotNeeded]) {
-                            log.info("[RESPONSE] [$issue] [${response.first}] Operation not needed")
+                            if (doLog) log.info("[RESPONSE] [$issue] [${response.first}] Operation not needed")
                         }
                         is FailedModuleResponse -> {
                             addFailedTicket(issue)
 
                             for (exception in it.exceptions) {
-                                log.error("[RESPONSE] [$issue] [${response.first}] Failed", exception)
+                                if (doLog) log.error("[RESPONSE] [$issue] [${response.first}] Failed", exception)
                             }
                         }
                     }
                 }, {
-                    log.info("[RESPONSE] [$issue] [${response.first}] Successful")
+                    if (doLog) log.info("[RESPONSE] [$issue] [${response.first}] Successful")
                 })
             }
 
@@ -110,7 +111,7 @@ class ModuleExecutor(
             .filterValues { it.isLeft() }
             .forEach { entry ->
                 (entry.value as Either.Left).a.exceptions.forEach {
-                    log.error("[UPDATE] [${entry.key}] Failed", it)
+                    if (doLog) log.error("[UPDATE] [${entry.key}] Failed", it)
                 }
                 addFailedTicket(entry.key)
             }
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -20,8 +20,7 @@ class ModuleExecutor(
     private val queryCache: Cache<List<Issue>>,
     private val issueUpdateContextCache: IssueUpdateContextCache,
     private val searchIssues:
-        (Cache<MutableSet<String>>, Cache<MutableSet<String>>, String, Int, () -> Unit) -> List<Issue>,
-    private val doLog: Boolean = true
+        (Cache<MutableSet<String>>, Cache<MutableSet<String>>, String, Int, () -> Unit) -> List<Issue>
 ) {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -63,7 +64,7 @@ class ModuleExecutor(
             } while (missingResultsPage)
             return ExecutionResults(true, failedTickets)
         } catch (ex: Throwable) {
-            log.error("Failed to execute modules", ex)
+            if (doLog) log.error("Failed to execute modules", ex)
```
urielsalis, line 67: Wouldn't this prevent logging in debug mode too? IIRC you can set conditions directly in logback.xml that affect only the discord appender

urielsalis:
https://gist.github.com/traviskaufman/d13b29807c0be730db45

That's how to disable the appender(by having the same configuration minus the appender) in tests

For local we can make sure it doesn't fail and not set a webhook

urielsalis:
While we are at it, I would disable the file appender in tests too

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/Arisa.kt
```diff
@@ -9,6 +9,7 @@ object Arisa : ConfigSpec() {
         val username by required<String>()
         val password by required<String>()
         val dandelionToken by required<String>(description = "Token for dandelion.eu")
+        val discordLogWebhook by required<String>(description = "Webhook to post log in a Discord channel")
```
urielsalis, line 12: Can't you make this optional instead of required so you don't need to put optional=true while reading the config?

SPGoding: Actually the `optional = true` part is in case there's no `arisa.yml` file while testing. But anyways the `readConfig` method won't be executed when testing so we can remove that part safely

src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,12 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="debug">
+        <appender-ref ref="STDOUT" />
+        <appender-ref ref="FILE" />
```
urielsalis, line 10:
```suggestion
```

src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
```diff
@@ -13,7 +13,8 @@ class IntegrationTest : StringSpec({
                 mapOf(
                     "arisa.credentials.username" to "test",
                     "arisa.credentials.password" to "test",
-                    "arisa.credentials.dandelionToken" to "test"
+                    "arisa.credentials.dandelionToken" to "test",
+                    "arisa.credentials.discordLogWebhook" to "test"
```
urielsalis, line 17: I would make it optional and test what happens when we dont include it(shouldnt log to discoerd if its not there)

NeunEinser:
If that's possible and easy to do, it would also be nice if all errors could be logged to a different channel additionally. In our use case, it would make sense to have an #arisa-log with the entire log, and post all errors to #arisa-development, so we can react quickly.

urielsalis:
I would argue that we should only log errors, as success messages are fine(and plently, we dont want to get ratelimited by discord)

urielsalis:
Its also easier to do, as for only logging errors you just add this to the appender

`    <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
        <level>ERROR</level>
    </filter>
`

NeunEinser:
I think there would still be value for a full log. For example, when a new module is added, it can be important to see if the tickets that module was successful on, were actually correct and executed correctly.

urielsalis:
Having two appenders with different channels?

NeunEinser:
Yes.
Also about rate limiting: Over the entire day of 2020-06-28 we had 306 non-error messages. that's only about 12.75 messages per hour. I don't think we should get any problems with that.
Usually, the most entries happen when there is an error that makes modules fail to executed. Yesterday's log was spammed with error messages due to a name being null somewhere.

src/main/resources/logback.xml
```diff
@@ -21,8 +21,23 @@
         </encoder>
     </appender>
 
+    <appender name="DISCORD" class="com.github.napstr.logback.DiscordAppender">
+        <webhookUri>https://discordapp.com/api/webhooks/937000047492844289/yoUKnOwwHAtTHiSiSFaKe</webhookUri>
```
Marcono1234, line 25: Would it be possible to not set this or leave it blank? `DiscordAppender` should then wait until it is set before trying to send messages, see:
https://github.com/napstr/logback-discord-appender/blob/a653e826872adb354d2180066ccc1812bca39ac8/src/main/java/com/github/napstr/logback/DiscordAppender.java#L57

urielsalis: thats what im asking in https://github.com/mojira/arisa-kt/pull/350#discussion_r448792562

src/test/kotlin/io/github/mojira/arisa/infrastructure/IntegrationTest.kt
SPGoding: `discordLogWebhook` is required

SPGoding:
> 
> 
> Yes.
> Also about rate limiting: Over the entire day of 2020-06-28 we had 306 non-error messages (and 2 errors when there was some kind of hickup on jira's side and Arisa failed to search issues). That's only about 12.75 messages per hour. I don't think we should get any problems with that.
> Usually, the most log entries happen when there is an error that makes modules fail to execute. Yesterday's log was spammed with error messages due to a name being null somewhere.

Afaik the appender seems to handle the rate limit automagically.

PR #358, opened 2020-07-08
Fix #354

### LINKED ISSUE #354: Resolving a ticket for having an unreleased version as AR takes precedence over resolving a ticket for any other reason
## The Bug
Arisa is resolving tickets for having and unreleased version as AR before checking if the ticket is in a different language or has an unauthorized version of the game and resolving them as Invalid. This allows the reporter to reopen their invalid tickets.
## Example Ticket
https://bugs.mojang.com/browse/MC-193506
https://bugs.mojang.com/browse/MC-193508
https://bugs.mojang.com/browse/MC-193498
https://bugs.mojang.com/browse/MC-193492
https://bugs.mojang.com/browse/MC-193588
## Expected behavior
Arisa should be checking for things to make the issue invalid so the reporter doesn't have to worry about adding an affected version and instead sees how the issue is invalid and possibly opens a new ticket

urielsalis:
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
@@ -130,8 +130,6 @@ class ModuleRegistry(private val config: Config) {
             "updated > ${checkStart.toEpochMilli()} AND updated < ${checkEnd.toEpochMilli()}"
         }
 
-        register(Modules.FutureVersion, FutureVersionModule(config[Modules.FutureVersion.message]))
-
         register(Modules.HideImpostors, HideImpostorsModule())
 
         register(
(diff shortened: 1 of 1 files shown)


SPGoding:
> @SPGoding As it looks like, the execution is applied per module, and not after the run for all changes.
> 
> This is in executeModule and is thus done per module:
> https://github.com/mojira/arisa-kt/blob/8d068d0d8ce44dc2efb03442a085047d28c535cf/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt#L108-L116
> 
> So, having the module earlier means it takes precedence.

Oops you're right. Forgot the fact that we moved the apply changes to the end of each module execution instead of each run :facepalm:

Issue #360, opened 2020-07-11
Arisa should disregard or delete modded crash reports which not attached by the reporter


violine1101:
## The Bug
Arisa currently closes a ticket as Invalid if _anyone_ attaches a modded crash report.

## Example Ticket
https://bugs.mojang.com/browse/MC-173299

## Expected behavior
Either delete the modded crash report if it's not been attached by the reporter, or simply ignore it.

NeunEinser:
I don't think the problem in your example was that someone other than the reporter attached a crash. The problem was that the bot did close the issue, even though one of the attached reports was not modded.

PR #361, opened 2020-07-12
Improve crash module

NeunEinser:
## Purpose
This fixes some issues the crash module had so far. The idea comes from #360, although this does not require the crash report to have been uploaded by the reporter.

## Approach
The crash module now requires the following:
* One crash report was uploaded since the last run.
* All of the uploaded crash reports are invalid (so either a dupe or modded). It does not matter when those were uploaded.

Since the first point already requires a recent report and the 2nd one requires all reports to be accessible, I removed the configurable option of maxAttachmentAge.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -189,7 +189,6 @@
       },
       "crash": {
         "excludedStatuses": ["Postponed"],
-        "maxAttachmentAge": 30,
         "crashExtensions": [
           "txt",
           "log"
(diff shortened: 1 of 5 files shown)


NeunEinser:
@violine1101 Do you think these changes would be sufficient for the sake of #360?

PR #363, opened 2020-07-12
Only transfer to newly added parents

### LINKED ISSUE #362: Users can add affected versions to reports they made duplicates of
## The Bug
If a user adds an affected version to their resolved duplicate ticket, Arisa automatically adds that affected version to the new report.
This can be used by random users to add affected versions to Fixed tickets or add random versions to tickets that aren't theirs
## Example Ticket
https://bugs.mojang.com/browse/MC-193312 duplicate of https://bugs.mojang.com/browse/MC-191016
## Expected behavior
Users shouldn't be able to add these affected versions to tickets using their duplicates.

NeunEinser:
## Purpose
Fix #362 

## Approach
Transfer modules now only transer to newly added links, so changes made after the link was added will not be considered.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


NeunEinser:
Merging now, since there is opportunity to abuse this with the new version again now.

PR #370, opened 2020-07-15
Fix #356

### LINKED ISSUE #356: If a transfer module fails, last run saves the parent ticket instead of the ticket where Arisa tried to transfer from
## The Bug
The transfer modules (currently TransferLinks and TransferVersions) work on Duplicated issues and will only see the Duplicate report. So if they fail, the Duplicate they fail on should be saved in last-run to be executed again on the next restart after a potential bug fix.
However, the update of course needs to happen in the parent ticket(s). In the current infrastructure, the modules themselves don't fail but the update fails, and the ticket Arisa tried to update gets stashed in last-run instead because of this.

## Example Ticket
Arisa failed on [MC-180306](https://bugs.mojang.com/browse/MC-180306) (see #355) and last run contained [MC-150820](https://bugs.mojang.com/browse/MC-150820) instead.

## Expected behavior
Save the ticket the module was working on instead of the ticket the update failed on.
This could mean introducing an additional optional parameter to the update context. Since multiple modules can potentially update the same ticket, the context should contain a set of origin tickets, if that solution ends up being implemented.

urielsalis:
## Purpose
Fix #356
## Approach
If we are removing all versions, resolve as awaiting response. Else just put comment
#### Checklist
- [x] Included tests (Bad ones)
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
@@ -25,7 +25,9 @@ class FutureVersionModule(
 
             latestVersion!!.add()
             removeFutureVersions.forEach(::run)
-            resolveAsAwaitingResponse()
+            if (affectedVersions.size <= removeFutureVersions.size) {
+                resolveAsAwaitingResponse()
+            }
             addComment(CommentOptions(message))
         }
     }
--- a/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
@@ -9,7 +9,8 @@ import io.github.mojira.arisa.domain.Version
 import java.time.Instant
 
 class FutureVersionModule(
-    private val message: String
+    private val messageFull: String,
+    private val messagePanel: String
 ) : Module {
     override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
         Either.fx {
@@ -25,10 +26,12 @@ class FutureVersionModule(
 
             latestVersion!!.add()
             removeFutureVersions.forEach(::run)
-            if (affectedVersions.size <= removeFutureVersions.size) {
+            if (affectedVersions.size > removeFutureVersions.size) {
+                addComment(CommentOptions(messagePanel))
+            } else {
                 resolveAsAwaitingResponse()
+                addComment(CommentOptions(messageFull))
             }
-            addComment(CommentOptions(message))
         }
     }
 
--- a/src/test/kotlin/io/github/mojira/arisa/modules/FutureVersionModuleTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/modules/FutureVersionModuleTest.kt
@@ -276,6 +276,22 @@ class FutureVersionModuleTest : StringSpec({
         result.shouldBeRight(ModuleResponse)
     }
 
+    "should not resolve if there is another version" {
+        val module = FutureVersionModule("message")
+        val issue = mockIssue(
+            created = FIVE_SECONDS_AGO,
+            affectedVersions = listOf(FUTURE_VERSION, RELEASED_VERSION),
+            changeLog = listOf(ADD_FUTURE_VERSION),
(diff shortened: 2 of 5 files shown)


src/test/kotlin/io/github/mojira/arisa/modules/FutureVersionModuleTest.kt
```diff
@@ -276,6 +276,22 @@ class FutureVersionModuleTest : StringSpec({
         result.shouldBeRight(ModuleResponse)
     }
 
+    "should not resolve if there is another version" {
+        val module = FutureVersionModule("message")
+        val issue = mockIssue(
+            created = FIVE_SECONDS_AGO,
+            affectedVersions = listOf(FUTURE_VERSION, RELEASED_VERSION),
```
urielsalis, line 283: This doesnt really test the behavour(as there is no easy way after the caching work), but it serves as a spec of what themodule does

src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
```diff
@@ -25,7 +25,9 @@ class FutureVersionModule(
 
             latestVersion!!.add()
             removeFutureVersions.forEach(::run)
-            resolveAsAwaitingResponse()
+            if (affectedVersions.size <= removeFutureVersions.size) {
+                resolveAsAwaitingResponse()
+            }
             addComment(CommentOptions(message))
```
NeunEinser, line 31: The message also should change. The message that is used always with this pr suggests that the ticket has been ARed
```
{panel:borderColor=orange}(!) Please do not mark _Unreleased Versions_ as affected. You don't have access to them yet.{panel}

_We have removed it and added the latest released version_

Please edit the *Affects Version/s* field and select the specific *Released Version* you were using when you encountered the reported issue.

If you can't find it in the list, please make sure that:
- You are in the *correct project* on the bug tracker.
- You were playing the *latest release version* or the *latest development version* of the game. 

~This issue has been temporarily closed as {color:#d04437}*Awaiting Response*{color}. Once the requested information has been delivered, the report will be reopened automatically.~

*Quick Links*:
📓 [Issue Guidelines|https://bugs.mojang.com/projects/MC/summary] -- 💬 [Community Support|https://discord.gg/58Sxm23] -- 📧 [Customer Support|https://help.minecraft.net/hc/en-us/requests/new] -- ✍️ [Feedback and Suggestions|https://feedback.minecraft.net/] -- 📖 [Game Wiki|https://minecraft.gamepedia.com/Minecraft_Wiki]
```

We do have the panel on its own and I suggest just usning the message called Panel - Future Version when not resolving as AR:
```
{panel:borderColor=orange}(!) Please do not mark _Unreleased Versions_ as affected. You don't have access to them yet.{panel}
```

src/test/kotlin/io/github/mojira/arisa/modules/FutureVersionModuleTest.kt
NeunEinser: You should be able to set the resolveAsAwaitingResponse function and assert that it is not being called, like we usually do.

SPGoding: Already done XD

src/test/kotlin/io/github/mojira/arisa/modules/FutureVersionModuleTest.kt
```diff
@@ -276,8 +282,28 @@ class FutureVersionModuleTest : StringSpec({
...
+
+        val module = FutureVersionModule("message", "panel")
+        val issue = mockIssue(
+            created = FIVE_SECONDS_AGO,
+            affectedVersions = listOf(FUTURE_VERSION, RELEASED_VERSION),
+            changeLog = listOf(ADD_FUTURE_VERSION),
+            project = mockProject(
+                versions = listOf(RELEASED_VERSION)
+            ),
+            resolveAsAwaitingResponse = { isResolved = true }
```
SPGoding, line 296: Looks like we could also test the comment here

NeunEinser: Probably should also add another test for testingthat the correct comment is posted in the other case

PR #371, opened 2020-07-15
Fix re-run the ticket being updated instead of the one trigger the update

### LINKED ISSUE #356: If a transfer module fails, last run saves the parent ticket instead of the ticket where Arisa tried to transfer from

SPGoding:
# Purpose
Fix #356
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
@@ -104,7 +104,7 @@ class ModuleExecutor(
                     log.info("[RESPONSE] [$issue] [${response.first}] Successful")
                 })
                 issueUpdateContextCache.storage.forEach {
-                    it.value.triggeredBy = issue
+                    it.value.triggeredBy = it.value.triggeredBy ?: issue
                 }
             }
 
@@ -114,7 +114,7 @@ class ModuleExecutor(
             .forEach { (updateTo, pair) ->
                 val (triggeredBy, result) = pair
                 (result as Either.Left).a.exceptions.forEach {
-                    log.error("[UPDATE] [TO ${updateTo}] [BY ${triggeredBy}] Failed", it)
+                    log.error("[UPDATE] [TO $updateTo] [BY $triggeredBy] Failed", it)
                 }
                 addFailedTicket(triggeredBy!!)
             }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -103,16 +103,20 @@ class ModuleExecutor(
...
+                }
             }
 
         issueUpdateContextCache.storage
-            .mapValues { applyIssueChanges(it.value) }
-            .filterValues { it.isLeft() }
-            .forEach { entry ->
-                (entry.value as Either.Left).a.exceptions.forEach {
-                    log.error("[UPDATE] [${entry.key}] Failed", it)
+            .mapValues { Pair(it.value.triggeredBy, applyIssueChanges(it.value)) }
```
urielsalis, line 112: Do we have a test for the cache in the moduleexecutor?

SPGoding: We have one for adding failed tickets, however that's for the module response section instead of the issue update section. So afaik we don't have tests for that, though I might miss something.

I was trying to add a test for this but wasn't sure how to do it

SPGoding:
I gave up on testing

SPGoding:
I have no curly braces on that line, waht are you stupid lint talking about

NeunEinser:
Late to the party, but I reviewed and approve :P

PR #374, opened 2020-07-17
Add new check in pr template

SPGoding:
This option is mostly used for reminding the person to update the Wiki

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,5 +6,5 @@
 
 #### Checklist
 - [ ] Included tests
-- [ ] Edited documentation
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -6,4 +6,5 @@
 
 #### Checklist
 - [ ] Included tests
+- [ ] Edited documentation
```
Marcono1234, line 9: Maybe it would be good to mention where the documentation is. So what do you think about the following?
```suggestion
- [ ] Updated documentation in [README](https://github.com/mojira/arisa-kt/blob/master/README.md) and [wiki](https://github.com/mojira/arisa-kt/wiki)
```

PR #393, opened 2020-07-29
Add ARISA_PURGE_ATTACHMENT command

### LINKED ISSUE #306: Delete accidentally uploaded unzipped worldfiles
## The Problem
Occasionally, it happens that a user uploads a world by dragging the world folder into the issue. That will cause all world files to be uploaded separately.

## Example Ticket
[MC-161823](https://bugs.mojang.com/browse/MC-161823)

## Possible Solutions
1. Add *.mca and *.dat to the attachment blacklist
This could have unwanted side effects though. It might be useful to upload a level.dat file sometimes, for example.
This could be solved by explicitly allowing level.dat somehow, however that would require an extension of the module
2. Command module for deleting attachments
Another soluting could be a command module that either purges all attachments from a ticket, or takes an extension as parameter.

SPGoding:
## Purpose
Fix #306.
## Approach
By adding a new `ARISA_PURGE_ATTACHMENT <start ID> [<stop ID>]` command.
#### Checklist
- [x] Included tests
- [x] Tested in [MCTEST-98](https://bugs.mojang.com/browse/MCTEST-98) and [MC-196464](https://bugs.mojang.com/browse/MC-196464)

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -187,6 +187,20 @@
         "excludedStatuses": ["Postponed"],
         "message": "incomplete"
       },
+      "command": {
+        "resolutions": [
+          "Awaiting Response",
+          "Cannot Reproduce",
+          "Done",
+          "Duplicate",
+          "Fixed",
+          "Incomplete",
+          "Invalid",
+          "Unresolved",
+          "Won't Fix",
+          "Works As Intended"
+        ]
+      },
       "crash": {
         "excludedStatuses": ["Postponed"],
         "crashExtensions": [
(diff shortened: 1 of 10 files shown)


arisa.json
```diff
@@ -187,6 +187,20 @@
         "excludedStatuses": ["Postponed"],
         "message": "incomplete"
       },
+      "command": {
```
urielsalis, line 190: Are we sure we want to limit the entire command module to those resolution?

Issue #182, opened 2020-05-03
RevokeFieldChangesModule: Ignore old Mojang Priority "Normal" changelog entries


violine1101:
## The Bug
Originally, all tickets had Mojang Priority set as Normal. However, that was changed when the workflow was changed from "Global Mojang Workflow" to "Java Workflow", as the default value for the Mojang Priority field was now EMPTY. There was no changelog entry for this change (see https://bugs.mojang.com/browse/MC-123693 for example).

Arisa incorrectly resets the priority to "Normal" on tickets where this happened.

## Example Ticket
https://bugs.mojang.com/browse/MC-11739

## Expected behavior
Arisa should ignore all changelog entries that set "Mojang Priority" to Normal if they were before June 5.

urielsalis:
I dont think the module exists anymore. Is this still an issue?

NeunEinser:
It is something we still must consider in #195.
The temporary removal of a module is not an acceptable resolution for an issue, but if you want you can add it in a comment in that pr for reference instead.

chandler05:
I'm going to close this as #195 was closed, and because users can no longer update the Mojang Priority, so this is no longer an issue

PR #408, opened 2020-08-05
Check resolved tickets for future versions

### LINKED ISSUE #405: Arisa does not remove Future Versions from resolved tickets
## The Bug
When a user adds a future version to a resolved ticket, the version is not removed
## Example Ticket
Do it yourself ;)

## Expected behavior
It should be removed because it is an unreleased/future version

chandler05:
Fix #405 
(I hope I did this correctly, its my first pull request)

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
@@ -31,6 +31,8 @@ class FutureVersionModule(
                 if (resolution == null || resolution == "Unresolved") {
                     resolveAsAwaitingResponse()
                     addComment(CommentOptions(messageFull))
+                } else {
+                    addComment(CommentOptions(messagePanel))
                 }
             }
             removeFutureVersions.forEach(::run)
--- a/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
@@ -29,9 +29,6 @@ class FutureVersionModule(
             } else {
                 latestVersion!!.add()
                 if (resolution == null || resolution == "Unresolved") {
-                                        resolveAsAwaitingResponse()
-                }
-                addComment(CommentOptions(messagePanel))
                     resolveAsAwaitingResponse()
                     addComment(CommentOptions(messageFull))
                 } else {
--- a/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
@@ -20,6 +20,18 @@ class FutureVersionModule(
                 .filter { it.id in addedVersions }
                 .map { it.remove }
             assertNotEmpty(removeFutureVersions).bind()
+            assertEither(
+                assertEquals(resolution, "Awaiting Response",),
+                assertEquals(resolution, "Cannot Reproduce"),
+                assertEquals(resolution, "Done"),
+                assertEquals(resolution, "Duplicate"),
+                assertEquals(resolution, "Fixed"),
+                assertEquals(resolution, "Incomplete"),
+                assertEquals(resolution, "Invalid"),
+                assertEquals(resolution, "Unresolved"),
+                assertEquals(resolution, "Won't Fix"),
+                assertEquals(resolution, "Works As Intended")
+            ).bind()
 
             val latestVersion = project.versions.lastOrNull(::isFutureVersion.complement())
             assertNotNull(latestVersion).bind()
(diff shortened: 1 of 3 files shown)


chandler05:
I have read the CLA Document and I hereby sign the CLA

chandler05:
I have read the CLA Document and I hereby sign the CLA

src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
```diff
@@ -20,6 +20,18 @@ class FutureVersionModule(
                 .filter { it.id in addedVersions }
                 .map { it.remove }
             assertNotEmpty(removeFutureVersions).bind()
+            assertEither(
```
urielsalis, line 23: If I remember correctly, you can do this in arisa.json instead :D

urielsalis: as a note for the future(not needed if you do it in arisa.json), you should add tests to each change

chandler05: oh gosh I don't understand what that means :|

chandler05:
What did I mess up?

urielsalis:
Tests are failing, probably because they use one of the blacklisted extensions

You can instead do something like the attachment module https://github.com/mojira/arisa-kt/blob/master/arisa.json#L38 and do it via config

chandler05:
I switched it to arisa.json, hopefully that fixes it

chandler05:
YASSS IT DID IT

SPGoding:
Great job!

chandler05:
Is it fine if a message is not given when adding a future version to a resolved ticket?

chandler05:
Otherwise I can just add `else {
                    addComment(CommentOptions(messagePanel))
                }` into the module

chandler05:
What normally happens first, reopening a ticket resolved as AR, or resolving a ticket as AR for Future Versions?

SPGoding:
Resolves always happen at last, no matter what you write in the code.

chandler05:
So if somebody added a future version to their AR ticket, it would be reopened, and then the future version would be removed and the ticket reresolved?

SPGoding:
The status of a ticket doesn't update in the same run. So if somebody added a future version to their AR ticket, what really happens is:
- It would be marked as _should_ reopen by the ReopenAwaiting module.
- The future version would be marked as _should_ remove by the FutureVersion module. Because the ticket is still resolved, FutureVersion won't try re-resolve this ticket.
- ...
- Actually reopened the ticket
- Actually removed the future version
- Finished

chandler05:
So should AR be added to the if statement with Null and Unresolved?

SPGoding:
_Actually we have a backlogged PR working on a general RevokeFieldModule: #195_

Also this module doesn't seem to consider #182.

chandler05:
But that module was setting ticket's priorities to Normal instead of None because of the changes done to the workflow. My module's not setting it to Normal

chandler05:
As for the backlogged PR, this could be considered a temporary module in place of that until that one is done, but its fine if this module isn't used 🤷

chandler05:
SPGoding your comment was in the wrong pull request :P

src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
```diff
@@ -28,7 +28,7 @@ class FutureVersionModule(
                 addComment(CommentOptions(messagePanel))
             } else {
                 latestVersion!!.add()
-                if (resolution == null || resolution == "Unresolved") {
+                if (resolution == null || resolution == "Unresolved" || resolution == "Awaiting Response") {
```
SPGoding, line 31: This is really hacky and I don't think we want to do so. Also there are cases where the ReopenAwaiting module won't reopen the ticket even if a version is added, e.g. the version is added by a volunteer.

chandler05: So what should be done to prevent both cases from happening?

src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt
```diff
@@ -28,8 +28,12 @@ class FutureVersionModule(
                 addComment(CommentOptions(messagePanel))
             } else {
                 latestVersion!!.add()
-                resolveAsAwaitingResponse()
-                addComment(CommentOptions(messageFull))
+                if (resolution == null || resolution == "Unresolved") {
```
urielsalis, line 31:
```suggestion
                if (resolution == null || resolution == "Unresolved") {
                                        resolveAsAwaitingResponse()
                }
                addComment(CommentOptions(messagePanel))
```

urielsalis:
Looks good to me. @SPGoding?

PR #414, opened 2020-08-18
Restrict command comments to proper groups

### LINKED ISSUE #404: Command feedback comment for staff restricted commands should be staff restricted
## The Problem

## Possible Solution

chandler05:
## Purpose
Fix #404
## Approach
I created a var that determines whether or not the command is mod restricted
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
@@ -20,6 +20,7 @@ class CommandModule(
     val purgeAttachmentCommand: Command = PurgeAttachmentCommand(),
     val deleteCommentsCommand: Command = DeleteCommentsCommand()
 ) : Module {
+    var modRestricted = false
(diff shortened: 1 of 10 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
```diff
@@ -53,14 +62,20 @@ class CommandModule(
         return when (split[0]) {
             // TODO this should be configurable if we move to a registry
             // TODO do we want to add the response of a module via editing the comment?
-            "ARISA_ADD_VERSION" -> addVersionCommand(issue, *arguments)
+            "ARISA_ADD_VERSION" -> {
+                modRestricted = false
```
urielsalis, line 66: Same as the other PR. Its bad practice to store state globally to controlthis, plus multiple commands will all do whatever the last command was. You can encode this on the return of the module (Have a data class for both username and the boolean for its restricted?)

Issue #413, opened 2020-08-14
Commands: Don't fail silently when command arguments are invalid


Marcono1234:
## The Problem
It appears currently all commands fail silently when the arguments are invalid:
- `ARISA_ADD_VERSION`:
  - missing argument
  - version already marked as affected
  - version does not exist
- `ARISA_FIXED`:
  - missing argument
  - _currently not checking if version exists?_
- `ARISA_PURGE_ATTACHMENT`:
  - too many arguments
  - arguments not integer
- `ARISA_REMOVE_COMMENTS`:
  - missing argument

This makes it difficult to spot mistakes and makes you think that Arisa for some reason did not recognize the command.

## Possible Solution
If possible it would be good to add a helper/staff restricted comment (with respective visibility, see also #404) describing why the command could not be executed, e.g. "Missing argument", "Version '_%s_' does not exist", ...

Alternatively we can edit the command comment to add the response (see the comment [here](https://github.com/mojira/arisa-kt/pull/416#issuecomment-677786139)).

PR #416, opened 2020-08-18
Add more info to command response

### LINKED ISSUE #404: Command feedback comment for staff restricted commands should be staff restricted

### LINKED ISSUE #415: Commands: Include command executor in response message
I think it would be useful to know who executed a command in the response message for the command

chandler05:
## Purpose
Close #415 and closes #404 
## Approach
I added a parameter checking for the comment author and adding it into the command message
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
@@ -6,7 +6,6 @@ import arrow.core.left
 import arrow.syntax.function.partially1
 import io.github.mojira.arisa.domain.Comment
 import io.github.mojira.arisa.domain.Issue
-import io.github.mojira.arisa.domain.User
 import io.github.mojira.arisa.modules.commands.AddVersionCommand
 import io.github.mojira.arisa.modules.commands.Command
 import io.github.mojira.arisa.modules.commands.DeleteCommentsCommand
@@ -15,6 +14,7 @@ import io.github.mojira.arisa.modules.commands.PurgeAttachmentCommand
 import java.time.Instant
 
 // TODO if we get a lot of commands it might make sense to create a command registry
+data class CommandInfo(val modRestricted: Boolean, val author: String, val commandName: String)
 class CommandModule(
     val addVersionCommand: Command = AddVersionCommand(),
     val fixedCommand: Command = FixedCommand(),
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
```diff
@@ -47,9 +49,11 @@ class CommandModule(
     }
 
     @Suppress("SpreadOperator")
-    private fun executeCommand(comment: String, issue: Issue, userIsMod: Boolean): Either<ModuleError, ModuleResponse> {
+    private fun executeCommand(comment: String, issue: Issue, userIsMod: Boolean, author: User):
+            Either<ModuleError, ModuleResponse> {
         val split = comment.split("\\s+".toRegex())
         val arguments = split.toTypedArray()
+        staffAuthor = author
```
urielsalis, line 56: lets not use global variables like this to pass this info. 

1- If there are multiple commands you will grab the username of latest for all
2- Its bad practice

You can modify the return of the either to include the author username(no need forthe entire object, just get their name)

src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
```diff
@@ -30,15 +32,15 @@ class CommandModule(
...
 
             when {
                 results.any { it.isLeft() && (it as Either.Left).a is FailedModuleResponse } -> {
-                    addRawRestrictedComment("Command execution failed", "helper")
+                    addRawRestrictedComment("$staffAuthor Command execution failed", "helper")
                     results.first { (it as Either.Left).a is FailedModuleResponse }.bind()
                 }
                 results.any { it.isRight() } -> {
-                    addRawRestrictedComment("Command execution was successful", "helper")
+                    addRawRestrictedComment("$staffAuthor Command execution was successful", "helper")
```
Marcono1234, line 43: Maybe it would be good to include a colon (`:`) after the executor name to structure the message a little bit more.

Marcono1234 (commented):
Would it also make sense to include the command name (though maybe not the arguments?) to give the message a little bit more context?
The output could then for example look like this:
> Marcono1234: ARISA_ADD_VERSION command execution was successful

SPGoding:
I'd prefer editing the command to add the result, so that both this and #414 will no longer be an issue.

Marcono1234:
> I'd prefer editing the command comment to add the result

That sounds like a good idea as well. Should the feedback then be added in front or behind the original command? I can imagine for more complex commands it would be useful to add it in front, e.g. after Arisa executed the command:
> Arisa: Execution succeeded / Execution failed
> <hr/>
> ARISA_ADD_VERSION 1.16

urielsalis:
Closing after discussion in discord, will be remade once we have brigadier

Marcono1234:
> Closing after discussion in discord, will be remade once we have brigadier

Should we have a separate GitHub issue for this? @SPGoding's idea of editing the command comment could also be combined with #413.

SPGoding:
Created #453 and updated #413 ;D

PR #417, opened 2020-08-20
Update Kotlin to v1.4

chandler05:
## Purpose
Update the Kotlin version to the newly released 1.4


urielsalis:
Make sure to use the migrate Kotlin version in telling, it will fix all problems for.you

Do note than when updating kotest you might need to also update the package names

chandler05:
It should be fine now though, correct? the builds are successful so I'd assume so

urielsalis (approved):

PR #422, opened 2020-08-28
Remove extra added versions to resolved tickets by users

### LINKED ISSUE #406: Remove affects versions added to resolved tickets depending on the resolution
## The Problem
Sometimes we get requests on discord to remove affects versions added by users to resolved tickets, and it would be nice to have this done automatically
## Possible Solution
If the ticket is resolved as Duplicate or Invalid (Maybe WAI, WF and Fixed too) and an affected version is added by a **Normal User** (Helpers and Mods should be able to add affected versions to resolved tickets) than Arisa should remove the affected version

chandler05:
## Purpose
Closes #406 
## Approach
Allows volunteers to add affected versions to resolved tickets, but removes them if added by a normal user

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
@@ -19,9 +19,7 @@ class RemoveVersionModule : Module {
 
     private fun Issue.getExtraVersionsLatelyAddedByNonVolunteers(lastRun: Instant): List<String> =
         if (created.isAfter(lastRun)) {
-            if (isVolunteer(reporter?.getGroups?.invoke())) {
-                emptyList()
-            } else if (resolution == "Unresolved") {
+            if (resolution == "Unresolved" || isVolunteer(reporter?.getGroups?.invoke())) {
                 emptyList()
             } else {
                 affectedVersions.map { ver -> ver.id }
@@ -30,6 +28,8 @@ class RemoveVersionModule : Module {
             changeLog
                 .asSequence()
                 .filter { it.created.isAfter(lastRun) }
+                .filter { it.field.toLowerCase() == "resolution" }
+                .filterNot { it.changedToString == "Unresolved" }
                 .filter { it.field.toLowerCase() == "version" }
                 .filterNot { isVolunteer(it.getAuthorGroups()) }
                 .mapNotNull { it.changedTo }
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
@@ -19,7 +19,7 @@ class RemoveVersionModule(
     }
 
     private fun Issue.getExtraVersionsLatelyAddedByNonVolunteers(lastRun: Instant): List<String> =
-        if (created.isAfter(lastRun) && (resolution == "Invalid" || resolution == "Won't Fix" || resolution == "Works As Intended" || resolution == "Duplicate" || resolution == "Incomplete")) {
+        if (created.isAfter(lastRun)) {
             if (isVolunteer(reporter?.getGroups?.invoke())) {
                 emptyList()
             } else {
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
@@ -19,7 +19,7 @@ class RemoveVersionModule : Module {
 
     private fun Issue.getExtraVersionsLatelyAddedByNonVolunteers(lastRun: Instant): List<String> =
         if (created.isAfter(lastRun)) {
-            if (isVolunteer(reporter?.getGroups?.invoke()) || resolution == "Unresolved") {
+            if (resolution == "Unresolved" || isVolunteer(reporter?.getGroups?.invoke())) {
                 emptyList()
             } else {
                 affectedVersions.map { ver -> ver.id }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -0,0 +1,39 @@
...
+            if (isVolunteer(reporter?.getGroups?.invoke())) {
+                emptyList()
+            } else {
+                affectedVersions.map { ver -> ver.id }
+            }
+        } else {
+            changeLog
+                .asSequence()
+                .filter { it.created.isAfter(lastRun) }
+                .filter { it.field.toLowerCase() == "version" }
```
Marcono1234, line 32: It appears even in Kotlin `toLowerCase()` uses the default locale so it would be good to specify `Locale.ENGLISH` / `Locale.ROOT`

chandler05: That's how the FutureVersionModule did it, and I copied it from there

chandler05: Added it anyways

chandler05: It seems that I can't do it. And since it's not completely necessary due to other modules using these functions without using the locale, I'm just not going to do it

Marcono1234: Hmm, yes it appears the project uses everywhere else `toLowerCase()` without locale as well. At least in Java you have to (sadly) always specify the locale since the default locale could have special lower casing rules (e.g. Turkish). Though it is unlikely that Arisa runs on a system with such a default locale, it is good to be on the safe side nonetheless.
Maybe @urielsalis (who seems to have much experience with Kotlin) knows if specifying a locale is necessary for Kotlin as well? Though it appears to be the case.

urielsalis: Generally yes, but we control the server(and we are about to put it on docker) and it's always set to UTC and English

Somewhere in the future we should probably add it everywhere

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -0,0 +1,38 @@
...
+                .filterNot { isVolunteer(it.getAuthorGroups()) }
+                .mapNotNull { it.changedTo }
+                .toList()
+        }
+
+    private fun isVolunteer(groups: List<String>?) = "helper" in (groups ?: emptyList())
```
NeunEinser, line 37:
```suggestion
    private fun isVolunteer(groups: List<String>?) = groups?.any { it == "helper" || it == "global-moderators" || it == "staff" } ?: false
```
This is how other modules do it. While global moderators is not strictly required currently, staff definitely is, since not all Mojang employees are in the helper group.

NeunEinser (changes requested):

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -0,0 +1,38 @@
...
+        if (created.isAfter(lastRun)) {
+            if (isVolunteer(reporter?.getGroups?.invoke())) {
+                emptyList()
+            } else {
+                affectedVersions.map { ver -> ver.id }
+            }
+        } else {
+            changeLog
+                .asSequence()
+                .filter { it.created.isAfter(lastRun) }
```
NeunEinser, line 30: I think it would be better additionally to this, to also check for after the issue's resolved date. If Arisa has a longer downtime, there could be a case where a version is added, and the issue is resolved afterwards.

src/test/kotlin/io/github/mojira/arisa/modules/RemoveVersionModuleTest.kt
```diff
@@ -0,0 +1,221 @@
...
+import io.kotest.assertions.arrow.either.shouldBeLeft
+import io.kotest.assertions.arrow.either.shouldBeRight
+import io.kotest.core.spec.style.StringSpec
+
+private val TWO_SECONDS_AGO = RIGHT_NOW.minusSeconds(2)
+private val FIVE_SECONDS_AGO = RIGHT_NOW.minusSeconds(10)
+
+private val VERSION = mockVersion(id = "2", released = true, archived = false)
+
+private val ADD_VERSION = mockChangeLogItem(field = "Version", changedTo = "1")
```
NeunEinser, line 17: These definitions are the cause for your failed tests. Your version that's in affectedVersions has id 2, while the changelog entry suggests a version with id 1 was added.

You should look into why the tests didn't run properly on your machine.

NeunEinser (approved):

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -0,0 +1,39 @@
...
+                .filter { it.id in addedVersions }
+                .map { it.remove }
+            assertNotEmpty(removeAddedVersions).bind()
+            removeAddedVersions.forEach(::run)
+        }
+    }
+
+    private fun Issue.getExtraVersionsLatelyAddedByNonVolunteers(lastRun: Instant): List<String> =
+        if (created.isAfter(lastRun)) {
+            if (resolution == "Unresolved" || isVolunteer(reporter?.getGroups?.invoke())) {
```
NeunEinser, line 22: You don't need to check the resolution here. The module doesn't see any tickets that do not match the resolution as defined in the json config.

chandler05: 👍

urielsalis (approved):

PR #425, opened 2020-09-04
Add a Remove Version Command

chandler05:
## Purpose
You can't remove archived versions from a ticket, and this fixes that. (Could be used to undo accidental adding affected versions through the add version command)
## Approach
Copied add version command and tweaked it.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveVersionCommand.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveVersionCommand.kt
@@ -1,0 +1,19 @@
+package io.github.mojira.arisa.modules.commands
+
+import arrow.core.Either
+import arrow.core.extensions.fx
+import io.github.mojira.arisa.domain.Issue
+import io.github.mojira.arisa.modules.ModuleError
+import io.github.mojira.arisa.modules.ModuleResponse
+import io.github.mojira.arisa.modules.assertTrue
+
+class RemoveVersionCommand : Command {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveVersionCommand.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.github.mojira.arisa.modules.ModuleError
+import io.github.mojira.arisa.modules.ModuleResponse
+import io.github.mojira.arisa.modules.assertTrue
+
+class RemoveVersionCommand : Command {
+    override fun invoke(issue: Issue, vararg arguments: String): Either<ModuleError, ModuleResponse> = Either.fx {
+        assertTrue(arguments.size > 1).bind()
+        val version = arguments.asList().subList(1, arguments.size).joinToString(" ")
+        assertTrue(issue.affectedVersions.any { it.name == version }).bind()
+        assertTrue(issue.project.versions.any { it.name == version }).bind()
```
Marcono1234, line 15: This line is probably not needed because it only matters whether the issue has the affected version. Additionally I doubt that an issue can have a version which does not exist as version for that project.

NeunEinser:
It is impossible to remove archived versions. At all. Even using the API.

This is why adding archived versions needs to be done with care, as it is mentioned in our helper workflow.

PR #428, opened 2020-09-08
Ability to modify multiple links

SunCatMC:
## Purpose
Adds ability to remove multiple links, and ability to both remove and add links from a list that isn't accepted by jira, mainly list with commas and full issue links

## Approach
Adds 2 commands: ARISA_ADD_LINKS and ARISA_REMOVE_LINKS that accept issue keys (case insensitive) or links in a list with spaces and/or commas

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-104


src/main/kotlin/io/github/mojira/arisa/modules/Helpers.kt
```diff
@@ -116,3 +118,112 @@ fun assertAfter(instant1: Instant, instant2: Instant) = if (instant1.isAfter(ins
...
     OperationNotNeededModuleResponse.left()
 }
+
+fun concatLinkName(vararg arguments: String): Array<String> {
+    val args = arguments.flatMap { s ->
+        s.split(',').filter {
+            it != ""
+        }
+    }.toTypedArray()
+    fun check(str: String): Boolean {
```
urielsalis, line 128: I would maybe move this function out of the other function, you can also turn it into a extension function like

fun String.checkIfLinkNameRegexMatches() or something like that(instead of str you use this)

src/main/kotlin/io/github/mojira/arisa/modules/Helpers.kt
```diff
@@ -116,3 +118,112 @@ fun assertAfter(instant1: Instant, instant2: Instant) = if (instant1.isAfter(ins
...
+    }
+    if (check(args[0])) {
+        val result = mutableListOf(*args)
+        result.add(0, "")
+        return result.toTypedArray()
+    }
+    var linkName = args[0]
+    var iSaved = 0
+    for (i in args.indices) {
+        iSaved = i
```
urielsalis, line 140: Im sure there is a function in kotlin to do this without a for loop

src/main/kotlin/io/github/mojira/arisa/modules/commands/DeleteLinksCommand.kt
```diff
@@ -0,0 +1,25 @@
...
+import io.github.mojira.arisa.modules.deleteLinks
+
+class DeleteLinksCommand : Command {
+    override fun invoke(issue: Issue, vararg arguments: String): Either<ModuleError, ModuleResponse> = Either.fx {
+        assertTrue(arguments.size > 2).bind()
+        val tmpArray = listOf(*arguments).subList(1, arguments.size).toTypedArray()
+        var args = concatLinkName(*tmpArray)
+        val type = args[0]
+        assertFalse(type == "").bind()
+        args = listOf(*args).subList(1, args.size).toTypedArray()
```
urielsalis, line 21: When you have to use var, its generally a bad idea 👀, Put this in another function 😅

SunCatMC: it happened 'cos I held everything in Array, which doesn't support kotlin list functions
I guess it would be better idea to pass lists as arguments then?

Otherwise I'm not sure how you would put that in a separate function <.<

urielsalis: If you are operating on it its probably better to pass as a list

SunCatMC: alright, gonna fix it then

SunCatMC: fixed by https://github.com/mojira/arisa-kt/pull/428/commits/6562b9f3985be0b05b318db5263c17b429cacee5

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -202,12 +202,18 @@ fun addRestrictedComment(
 }
 
 fun createLink(context: Lazy<IssueUpdateContext>, getContext: (key: String) -> Lazy<IssueUpdateContext>, linkType: String, linkKey: String, outwards: Boolean) {
-    context.value.otherOperations.add {
-        runBlocking {
-            Either.catch {
-                context.value.jiraIssue.link(linkKey, linkType)
+    when {
```
urielsalis, line 205: if :P

SunCatMC: facepalm.jpg
copied from TransferLinksModule tho

urielsalis: "Always leave the code better than you found it" :D

SunCatMC: fixed by https://github.com/mojira/arisa-kt/pull/428/commits/ace0b9230685c383f2e88d17e4e5a004c4a69f2a

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -201,13 +201,24 @@ fun addRestrictedComment(
...
+        context.value.otherOperations.add {
+            runBlocking {
+                Either.catch {
+                    context.value.jiraIssue.link(linkKey, linkType)
+                }
             }
         }
+    } else {
+        val key = context.value.jiraIssue.key
+        getContext(linkKey).value.jiraIssue.link(key, linkType)
```
SPGoding, line 221:
```suggestion
    } else {
        val key = context.value.jiraIssue.key
        createLink(getContext(linkKey), getContext, linkType, key, true)
```

By doing so the adding link operation for non-outwards links could also be added to the context's `otherOperations` ;D

SPGoding (approved):

urielsalis (approved):

Issue #421, opened 2020-08-28
Command for sending direct messages to users


Marcono1234:
## The Problem
In some cases it would be useful to send a direct message to a user instead of having to comment on an issue which informs all watchers, even if the message is not relevant for them.

## Possible Solution
Jira provides the API endpoint [/rest/api/2/issue/{issueIdOrKey}/notify](https://docs.atlassian.com/software/jira/docs/api/REST/7.6.1/#api/2/issue-notify) which can be used for this. This API endpoint sends an e-mail from the logged in account to users or groups.

### Considerations
- The e-mail is sent from Arisa, so it would be good to start the message with an indication which staff member created the message. However, it might also be useful to add an option for sending the message "anonymously" as "from staff" to prevent targeted attacks against staff members (depending on the context of the e-mail).
- The e-mail sent by Jira exposes the e-mail address of the user who used the API (i.e. Arisa) as `Reply-To` e-mail header. It would therefore be good to:
    - Not use a private e-mail address for Arisa
    - Not use Arisa's e-mail address for other purposes
    - Protect Arisa's account using [2FA](https://bugs.mojang.com/plugins/servlet/authplugin/tfaconfiguration) (?)

    Additionally since Arisa's e-mail address is used as `Reply-To` by Jira:
    - Arisa's mailbox should be checked a few times per week in case a user responds
    - The message sent by Arisa should have a footer stating
       > By responding to this e-mail your e-mail address will be shared with helpers and moderators of Mojira. They are volunteers and not Mojang Studios employees. Please only respond to this e-mail if you agree that your e-mail address is being shared with them. You can alternatively contact the helpers and moderators through the [Mojira Discord Server](https://discordapp.com/invite/rpCyfKV) or the [/r/Mojira subreddit](https://www.reddit.com/r/Mojira/) which both does not expose your e-mail address.
- Since the command becomes quite complex it would be good if Arisa edited the command message and added the execution feedback in front of the command (see also https://github.com/mojira/arisa-kt/pull/416#issuecomment-682845563).
- Before this feature is implemented it should be discussed with the Mojang devs, especially since they (likely) cannot trace these e-mails sent by the API. However, additionally the command could be configurable so it always sends a copy of the message to one (or multiple) of the Mojang developers by simply including them in the `users` value of the API request.
- When the command is used it should:
    - Send an e-mail to the staff member issuing the command (by including them in the `users` value of the API request) so they know that the command worked and see how the message looks to that user.
[...]

PR #432, opened 2020-09-14
Change platforms to multiple if duplicate tickets have a different platform (MEQS_KEEP_PLATFORM to prevent changes)

### LINKED ISSUE #430: Add MEQS_KEEP_PLATFORM to help prevent incorrect platform change / generalization
## The Problem
It is not always clear whether a bug is intended to be platform-specific.

## Possible Solution
If MEQS_KEEP_PLATFORM is set:

1, If someone changes the platform directly (usually to "Multiple") Arisa would either send a message to the helper/mod who did that, or revert the change and add a comment if was by a user.
2. If a duplicate is linked that has a different platform, the moderator making the link is notified.

Depends on Issue #421 being implemented.

### LINKED ISSUE #431: Automatically change platform to "multiple" if linked duplicate has different platform
## The Problem
Most bugs are not platform-specific, but the reporter should usually choose their own platform. It takes time to then look at the duplicates to see if they reference a different plaform

## Possible Solution
Automatically change platform to "multiple" unless MEQS_KEEP_PLATFORM is set (Issue #430) and a duplicate is linked with a different platform

chandler05:
## Purpose
Closes #431 and closes #430
## Approach
Copied and changed the ConfirmParentModule, and the KeepPrivateModule
Added in the Platform field


chandler05:
PR is ready for review 👍

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModule.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private fun isDuplicatedWithDifferentPlatforms(platform: String?, blacklist: List<String>, issue: Issue):
+            Either<ModuleError, Boolean> = Either.fx {
+        issue.links
+            .filter(::isDuplicatedLink)
+            .forEach {
+                val child = it.issue.getFullIssue().toFailedModuleEither().bind()
+                if (child.platform !in blacklist && child.platform != platform.getOrDefault("None")) {
+                    return@fx true
```
urielsalis, line 34: I don't think you need a FX block here, its really heavy on the call stack(3 levels causes a stackoverflow)

chandler05: That's how it's written in the ConfirrmParentModule

chandler05: That's where I took it from

urielsalis: maybe both shouldnt have it :P

urielsalis: the only reason is there is                 val child = it.issue.getFullIssue().toFailedModuleEither().bind()

You can just do an if for that (without toFailedModuleEither)

MikeUnverferth:
I think this should not be merged until issue #421 is implemented. Otherwise we won't know that there are unwanted platform conflicts on an issue with KEEP_PLATFORM set.

chandler05:
#421 is implemented is this PR already

chandler05:
That can most likely be handled separately. A lot of the modules already have similar consequences, and we've never had a command for direct messages. It just seems like it's not necessary in order to implement this PR, but it would be helpful in the future.

src/main/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModule.kt
chandler05: without doesn't exist? Plus, for some reason the platform and reporter fields are recognized by the code unless .bind() is attached to the end of the val child line

chandler05: I'm going to leave as-is right now unless its very necessary to do this

MikeUnverferth:
If there's already precedent for doing it that way then it's ok with me.

MikeUnverferth:
Could a message be added to arisa-error-log when it does not change the platform due to KEEP_PLATFORM being set? Though it's not technically an error, it's not a normal result either and it might get lost in the routine arisa-log.

urielsalis:
No. arisa-error-log is for errors we should act on(usually immideately), not for expected cases

src/main/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModule.kt
urielsalis: when I said without I mean removing .toFailedModuleEither() and just doing the if yourself

By the way, it doesnt really make sense to convert it to a either anyways

urielsalis:
And I do think we should add a warning message, but only if a volunteer/reporter tries to change it with the tag set(not on the duplicate thing). Same warning as in future version

arisa.json
```diff
@@ -175,6 +176,27 @@
...
+          "Windows 10 Mobile",
+          "Xbox One"
+        ],
+        "targetPlatform": ["Multiple"],
+        "transferredPlatformBlacklist": ["None"],
+        "keepPlatformTag": ["MEQS_KEEP_PLATFORM"],
+        "whitelist": ["MCPE"]
+      },
+      "keepPlatform": {
+        "keepPlatformTag": ["MEQS_KEEP_PLATFORM"],
```
urielsalis, line 197: Why is this a list? 🤔

chandler05:
I could never figure out how to add messages in, so that's why it doesn't have one

MikeUnverferth:
Can it write a [WARNING] entry to arisa-log? Same would be for whatever other "consequences" that other modules might do that a mod should be made aware of.

urielsalis:
A log is fine(info), but I really dont think its necessary

arisa.json
chandler05: Which part shouldn't be a list?

urielsalis: keepPlatformTag

chandler05: Is it? TargetPlatform is the same...
I... Uh, got it from the Arisa.json for ConfirmParent and KeepPrivate, and they're not lists but are written that way

urielsalis: keepPlatformTag": "MEQS_KEEP_PLATFORM" to make sure no one thinks we can put more than 1 (and read as String, not List<String>)

chandler05: Should I remove it from the target platform as well?

urielsalis: yep

MikeUnverferth:
If a mod resolves a PS4 ticket as a duplicate of an Xbox ticket with the KEEP_PLATFORM tag, that's either an incorrect resolution or the KEEP_PLATFORM tag was incorrectly set. In either case, a mod should look at the tickets and make an appropriate fix. Having a warning in the log will help to find these cases, though eventually sending a message directly would be better.

arisa.json
chandler05: K, removed from the targetConfirmationStatus as well

chandler05:
Or, the user could have put the wrong platform for the duplicate

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModule.kt
SunCatMC: getFullIssue() already returns an either, just a different one

and I don't see a cleaner way to deal with that honestly

and bind is needed 'cos when Either is Right it converts Either into its right half, that Boolean that you need

Issue #437, opened 2020-09-18
Remove embedded images from untrusted hosts


Marcono1234:
## The Problem
Mojira users can [embed images](https://bugs.mojang.com/secure/WikiRendererHelpAction.jspa?section=images) from arbitrary hosts. Not only can this make the page load content over an insecure connection (see also #436), but it might also expose data (such as IP address, browser, OS, ...) to a third party host because embedded images are loaded without requiring any interaction.

## Possible Implementation
Arisa should have a list of trusted hosts (e.g. `imgur.com`). When an issue or comment is created or edited, Arisa should check for any embedded images.

If an embedded image is:
- from a trusted host but uses HTTP: Arisa should change the URL to use HTTPS (would fix #436)
- from an untrusted host: Arisa should replace the embedded image with the URL to that image, e.g. `!example.com!` → `example.com`

## Considerations
Embedded image syntax parsing has to:
- ignore embedded images which are attached (i.e. image name is same as attachment name)
- consider that embedded image might specify attributes (e.g. `!https://example.com/image.png|align=right!`); these attributes should be ignored when parsing the host
- consider that URL apparently can also use IPv4 and IPv6 address as host

Jack-McKalling:
I think this issue is a real potential threat. But I do think we should ask for help assembling a list of trusted domains from mojang, or inquire one. Because they should be able to change the list.

Issue #385, opened 2020-07-26
Prevent normal users from changing the displayed text of a link by having Arisa always revert the link's text back to the link's url


chandler05:
## The Problem
As I stated in [WEB-2772](https://bugs.mojang.com/browse/WEB-2772), any user can link any text to any link. A user can pretend to be linking another ticket by editing the link text to an ticket number, when they actually are sending a malicious link or something similar.
## Possible Solution
Have Arisa revert any linked text to the url destination of the link from text created by a normal user. For example, when a user links to YouTube.com, but switches the text to say MC-12345, Arisa could change the text back to YouTube.com to show any user exactly where the link would be taking them.
I know this solution would stop Moderator and Helper links from looking the way they currently do, but I think that could be remedied by making so Arisa only reverts links from text created by non-mods and helpers, so the messages could stay the same.

Marcono1234:
The problem you are mentioning is definitely a valid one but the proposed solution might be a little bit too destructive. E.g. if someone comments:
> This is an intended feature which is [described|...] in the changelog.

Replacing "described" with the URL would break the sentence. Additionally if the URL is rather long it might decrease readability.

What about having Arisa instead inserting in superscript the domain behind the link, e.g.:
> This is an intended feature which is [described|...]^minecraft.net^ in the changelog.

Though maybe there are ways in which an adversary is able to circumvent this as well.

chandler05:
I'm not trying to solve the problem in WEB-2772, I'm just referring to it because there I explained how easy it was to make any link seem like a ticket ID/link

Pokechu22:
Browsers show the link target when you hover over it, and barring sites that do stupid javascript changes after you click it (google...) this information can prevent issues like that.  For mobile browsers, you can't hover, but it should still show the link if you long-press it (not as great though).

I think restricting link titles in general is an overreaction, but perhaps it'd be reasonable to detect links of the form `[https://some-other-url|https://a-url]` and maybe `[https://some-other-url|MC-1234]` where there is an expectation as to the target.

chandler05:
Actually, in Jira hovering over the link just says "Follow Link" instead of where it goes (At least in Chrome it does)

Pokechu22:
[...]

PR #438, opened 2020-09-21
Fix HTTP in links

### LINKED ISSUE #436: Change Imgur HTTP links to HTTPS
## The Problem
According to WEB-3002, imgur links sometimes load through HTTP instead of HTTPS. This causes the site to think that the user's connection is not secure.
## Possible Solution
Switch any Imgur links using HTTP to HTTPS to remove the warning

chandler05:
## Purpose
Closes #436 
## Approach
Replaces HTTP: in all links to HTTPS:


chandler05:
Is it fine to convert this for all links? Let me know if the HTTP needs to be changed only on imgur link, because right now it matches all Http: links

SPGoding:
I guess only a list of websites need to be replaced with HTTPS (see #437 for more information).

We could possibly maintain a list of trusted URLs in the config, create two regular expressions in the `ReplaceText` module from that list to solve the two issues mentioned in #437, and of course write a bunch of tests ;D

chandler05:
This could also potentially solve #385 by including a list of trusted links we no longer need to check for a link taking you to a bad website.

Anyone who's good at Regex feel free to contribute. It took me around 30 minutes to get the small Regex for this PR

List of trusted sites:
- Bug tracker (obviously)
- Jira/Atlassian (for WEB project)
- Reddit
- Discord
- Imgur
- YouTube
- Streamable
- Minecraft.net
- GitHub

Add others to the list! (I'll edit it with any new websites)

dericksonmark:
Hi! Maybe add OBS; it's in the helper message Attach video (PC). Also for Discord links, make sure you include discordapp.com and discord.gg, as they're both used in helper messages.

NeunEinser:
What do you mean with trusted URLs?
I don't think we need to restrict domains. Personally, I never encountered a bad link and this is probably gonna cause more harm than good.
The module that was proposed a while ago, which removes misleading urls is sufficient imo.

And yes, you shouldn't just replace http with https in general. Some sites still don't have the https protocol implemented, slicedlime.tv is an example of that. You can't connect using https, only http.

chandler05:
For now, I'm going to switch it to only Imgur links and then this PR can be reviewed/merged. The other module can be created separately

urielsalis (approved):

urielsalis:
Please open a new issue for adding more sites

PR #457, opened 2020-10-14
Fix tests not testing actual effects

### LINKED ISSUE #454: Some tests have missing checks if something has changed, making them succeed even though the module doesn't do what it's supposed to do
## The Bug
Some modules just check with `shouldBeRight ModuleResponse` at the end, without checking if the used functions has done anything, making it easy to make a passing module that is completely useless

Just having `shouldBe` inside passed functions doesn't help either as if you don't run the shouldBe (via not running the function it's inside), it's not gonna check for anything

## Example Module Tests
`AttachmentModuleTest` (there's "should return ModuleResponse when something is deleted successfully" but no tests that check whether the module removes anything)
`CHKModuleTest`
`EmptyModuleTest`
`HideImpostorsModuleTest`
`LanguageModuleTest`
`RemoveNonStaffMeqsModuleTest`
`RemoveTriagedMeqsModuleTest`
`RemoveVersionModule`
`ResolveTrashModule`

## Semi Affected Tests
* (can't come up with a way to reproduce the issue without triggering other tests, or only part of the functionality is faulty)

`CommandModuleTest` ("should return successfully when comment matches a command and it returns successfully")
`CrashModuleTest` (createLink checks are faulty, creation of comments is untested at all)
`FutureVersionModuleTest` (there are tests that expect success and expect some variable change, just not all)
`UpdateLinkedModuleTest` (only "should update if there is an old and a recent link" is missing a check)

## Expected behavior
The tests shouldn't be avoidable by faulty modules

Note: actually tested only in `CrashModule`, `RemoveNonStaffMeqsModule` and `RemoveTriagedMeqsModule`

SPGoding:
## Purpose
Fix #454 
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CommandModule.kt
@@ -45,7 +45,7 @@ class CommandModule(
                     addRawRestrictedComment("Command execution failed.\n~Author: $username~", "helper")
                     result.bind()
                 } else {
-                    addRawRestrictedComment("Command execution was successful\n~Author: $username~", "helper")
+                    addRawRestrictedComment("Command execution was successful.\n~Author: $username~", "helper")
                 }
             }
             ModuleResponse.right().bind()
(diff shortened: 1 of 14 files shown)


SunCatMC (approved):

SunCatMC:
I guess it can be merged, unless you want someone else to check your changes too

urielsalis (approved):

PR #465, opened 2020-10-18
Migrate several commands to CommandDispatcher

### LINKED ISSUE #404: Command feedback comment for staff restricted commands should be staff restricted

### LINKED ISSUE #413: Commands: Don't fail silently when command arguments are invalid

### LINKED ISSUE #453: Add brigadier as a dependency
## The Problem
Messy command registry.

## Possible Solution
Add brigadier as a dependency. See comment [here](https://github.com/mojira/arisa-kt/pull/416#issuecomment-696833252).

### LINKED ISSUE #532: Command feedback should be more detailed
## The Problem
There's no proper error reporting if a command fails. Example https://bugs.mojang.com/browse/MC-197177

Affects all commands probably.

## Possible Solution
Return proper error objects and build an error message from them.

SPGoding:
## Purpose
Resolve #453, fix #404, fix #413, and fix #532.
## Approach
- For #453, we added brigadier as a dependency.
- For #404, we changed the behavior from creating a new comment to editing the command result into the original comment.
- For #413, we added several command exceptions for illegal parameters. Thanks to brigadier, we also get a lot of detailed exceptions for syntax errors.
- For #532, ditto.

#### Checklist
- [x] Included tests
- [x] Tested in https://bugs.mojang.com/browse/MCTEST-116


urielsalis:
Love this, let me know when you are ready for it to be merged

urielsalis (approved):

PR #471, opened 2020-10-20
Remove references to whitelist and blacklist

SPGoding:
Disclaimer: I do not intend to start some sort of arguments here. I'm just providing these changes as a possible route that we could take. I believe whether this PR gets accepted or rejected doesn't make any of you/me/Mojira/whatever individual/community/organization/corporation more or less superior than others.

## Approach
Renamed references of `blacklisted` to `excluded` as the phrase `excludedStatuses` already existed in the code base before the changes.

Renamed most of references of `whitelisted` to `allowed` as it sounds the most natural.

#### Checklist
- [x] Updated tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -48,7 +48,7 @@
           "Won't Fix",
           "Works As Intended"
         ],
-        "extensionBlacklist": [
+        "excludedExtensions": [
           "jar",
           "exe",
           "0xe",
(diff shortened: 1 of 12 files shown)


SunCatMC:
the code seems fine, guess we should have a vote or something

arisa.json
```diff
@@ -48,7 +48,7 @@
           "Won't Fix",
           "Works As Intended"
         ],
-        "extensionBlacklist": [
+        "excludedExtensions": [
```
Marcono1234, line 51: "Excluded" is a little bit irritating because the purpose of the module is to remove attachments and the extensions here are not actually excluded but included for removal.
Maybe "disallowedExtensions" might be better?

Marcono1234:
The code of some modules and tests also uses the terms "blacklist" and "whitelist". It might be good to rename them as well to match their new config names (if we decide to merge in these changes).

SPGoding:
Closed due to lack of activity.

PR #485, opened 2020-11-21
Only allow the keep AR message once per ticket

### LINKED ISSUE #301: Add a delay to keep AR comments
## The Problem
When a ticket is kept in AR, Arisa should not reply to every single comment if there are many comments in a short amount of time.

Relates to, but is not the same as #190.

## Example Ticket
https://bugs.mojang.com/browse/MC-187083

## Possible Solution
Check if there have been any comments by arisa on a ticket recently (in the past day or so) and if there are, don't leave a comment. I don't know whether there's a feasible way to check whether the recently added comment(s) is the "no-reopen-ar" comment or not.

chandler05:
## Purpose
Fix #301
## Approach
Add a function to check for the keepAR message and don't comment again if one exists
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/ReopenAwaitingModule.kt
@@ -40,7 +40,9 @@ class ReopenAwaitingModule(
                 reopen()
             } else {
                 assertNotEquals(changeLog.maxByOrNull { it.created }?.author?.name, "arisabot")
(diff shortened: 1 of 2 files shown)


SPGoding (commented):
It would be great if you add a test for this.

chandler05:
Do you know what the body of the keepAR message is so I can apply it inside of the test?

chandler05:
oh wait, nevermind

SPGoding (approved):

PR #498, opened 2020-12-08
Exempt Mojang from the privacy module

### LINKED ISSUE #497: Exempt Mojang employees from the Privacy module
## The Problem
When a mojang employee makes a comment that contains an email address (as far as Arisa knows), the comment gets unnecessarily restricted, quite like here:
https://bugs.mojang.com/browse/MCPE-105882?focusedCommentId=858817&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-858817

## Possible Solution
We should be trusting the owner of the platform not to post privacy-sensitive information, so mojang employees should be exempted from the check. Alternatively, the other way around, the module could be configured to specifically function only on e.g. the user, helper and moderator groups. That way no exception-logic is needed.

chandler05:
## Purpose
Fix #497 
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
@@ -3,7 +3,6 @@ package io.github.mojira.arisa.modules
 import arrow.core.Either
 import arrow.core.extensions.fx
 import io.github.mojira.arisa.domain.CommentOptions
-import io.github.mojira.arisa.domain.Comment
 import io.github.mojira.arisa.domain.Issue
 import java.time.Instant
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -43,6 +43,7 @@ class PrivacyModule(
                 .filter { it.created.isAfter(lastRun) }
                 .filter { it.visibilityType == null }
                 .filter { it.body?.matches(patterns) ?: false }
+                .filterNot { it.getAuthorGroups()?.any { it == "staff" } ?: false }
```
SPGoding, line 46:
```suggestion
                .filterNot { it.getAuthorGroups()?.any { it == "helper" || it == "staff" } ?: false }
```

SPGoding (approved):

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
chandler05: I thought only Mojang employees were going to be exempt from this?

urielsalis: All helpers and mods should be too

urielsalis (commented):
.

PR #512, opened 2021-01-18
Remove all comments from user

urielsalis:
## Purpose
Remove all comments from certain users by issuing a command
## Approach
When command is triggered, look for the name in the arguments, call the jira api using the credentials to get their status, find all comments they ever did and set them to staff only
## Future work
Tested in TRASH-26834, needs unit tests, I ran out of time, but should work. Needs cleanup and tests


urielsalis:
To be cleaned up in #515

SPGoding (approved):

PR #516, opened 2021-01-26
Fix fetching user comments with spaces in user name

chandler05:
## Purpose
Fix fetching a user's comments if the user has a space in their username

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommand.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommand.kt
@@ -52,7 +52,8 @@ class RemoveUserCommand : Command {
                         .filter { it.author.name == name }
                         .forEachIndexed { index, it ->
                             it.update("Removed by Arisa - Delete user $name", "group", "staff")
-                            if (index % 10 == 0) {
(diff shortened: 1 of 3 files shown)


urielsalis:
Line 55 replace the 10 with a constant to fix the build

src/main/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommand.kt
```diff
@@ -52,7 +52,8 @@ class RemoveUserCommand : Command {
                         .filter { it.author.name == name }
                         .forEachIndexed { index, it ->
                             it.update("Removed by Arisa - Delete user $name", "group", "staff")
-                            if (index % 10 == 0) {
+                            const divisor = 10
```
urielsalis: have it at the top of the file, as const val 👁️

urielsalis: and in the JVM we use all caps for constants

urielsalis (approved):

urielsalis (approved):

Issue #510, opened 2021-01-15
Email addresses used by Mojang shouldn't be restricted


SPGoding:
## The Bug

Relates to #490 and #500.

## Example ticket
[https://bugs.mojang.com/browse/MC-211171?focusedCommentId=896562&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-896562](https://bugs.mojang.com/browse/MC-211171?focusedCommentId=896562&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-896562)

## Expected behavior
Don't restrict emails ending with `microsoft.com`, `minecraft.net`, `mojang.com`, and `zendesk.com`? We could use a regex to remove those domains from the text before checking, similar to what being suggested at #490.

urielsalis:
Merged into #541

PR #517, opened 2021-01-26
Update email regex in Privacy Module

### LINKED ISSUE #500: Only restrict email addresses of certain TLDs
## The Problem

Relates to #490.

`inventory_selected_icon_button@common.inventory_selected_icon_button` is seen as an email address.

## Possible Solution

Only restrict email addresses of certain common TLDs, e.g. `com`, `net`, `org`, `edu`, etc. I don't see the reason to add all new TLDs to the list (e.g. `xyz`, `io`, etc.).

### LINKED ISSUE #506: PrivacyModule: False negative if email address is in a link
## The Bug
The `PrivacyModule` doesn't trigger if the email address is in a link (example: `[test@example.com|mailto:test@example.com]`).

## Example Ticket
https://bugs.mojang.com/browse/MC-210895

## Expected behavior
The module should trigger.

chandler05:
## Purpose
Fix #506, Fix #500
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -12,7 +12,7 @@ class PrivacyModule(
 ) : Module {
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
-        """(?<!\[~)\b[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-_]{2,}\b(?!])""".toRegex(),
+        """[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-_]{2,}""".toRegex(),
```
SPGoding, line 15: Only removing the ending part `\b(?!])` should be able to fix this issue while still keeping the functionality of not hiding user mentions containing email addresses.

chandler05: Mentioning doesn't really do anything though, and once the user name has been changed from the email, then the mention will instead only display the raw email because that user name doesn't exist anymore, and arisa won't mark it as private

chandler05: Yeah, nevermind. You're probably right. Changed to just removing `\b(?!])`

SPGoding (approved):
It might be a good idea to also add a test for email in links?

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -12,7 +12,7 @@ class PrivacyModule(
 ) : Module {
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
-        """(?<!\[~)\b[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-_]{2,}\b(?!])""".toRegex(),
+        """(?<!\[~)\b[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-_]{2,}""".toRegex(),
```
urielsalis, line 15: This is kinda simple, we probably should work on a regex that only accepts valid TLDs at the end and use it on the privacy module too. @CubeTheThird maybe you have some ideas?

CubeTheThird: I might be missing some context, but the original pattern before the edit here doesn't allow for [ or ~ at the start, nor ] at the end. You've removed the check for the ] at the end, but the checks at the start are still present. What exactly is the goal then? Also, if you're looking at refining the detection of valid email addresses, there's a lot of info about that subject on https://emailregex.com/, but that's a bit of a separate conversation.

SPGoding: The goal of the regex is to match emails that should be hidden:
- `example@domain.com`
- `[email me|mailto:example@domain.com]`

But not emails in user mentions:
- `[~example.domain.com]`

SPGoding: It we are to improve the regex further, there are two open issues that could be considered: #500 and #510.

chandler05: I added #500 by limiting the number of charachters in the TLD and not matching with underscores according to RFC 1035 (underscores are not allowed in domain names)

urielsalis:
Is this ready to be merged?

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -12,7 +12,7 @@ class PrivacyModule(
 ) : Module {
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
-        """(?<!\[~)\b[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-_]{2,}\b(?!])""".toRegex(),
+        """(?<!\[~)\b[a-zA-Z0-9.-_]+@[a-zA-Z0-9.-_]+\.[a-zA-Z0-9.-^]{2,15}\b""".toRegex(),
```
Marcono1234, line 15: @chandler05 is it intended that `^` is allowed now in the top level domain regex? [RFC 952](https://tools.ietf.org/html/rfc952) does not mention that.

chandler05: Yes, it removes underscores from the regex as they aren't allowed in domains

Marcono1234: But why does it instead now add `^` as allowed character?

Marcono1234: Has been clarified, the regex range was added by accident (see also #539) and chandler wanted to preserve it.

PR #524, opened 2021-01-29
Encode username to UTF_8 when using the Remove User Command

chandler05:


urielsalis:
Just put "UTF-8"

chandler05:
I was going by this: https://stackoverflow.com/questions/10786042/java-url-encoding-of-query-string-parameters

urielsalis (approved):

PR #531, opened 2021-02-12
Add RemoveSpamModule

violine1101:
## Purpose
Remove spam following certain patterns as soon as it's posted.

## Approach
Monitor new comments and check if a regex matches at least `threshold` times.

## Future work
Hopefully keep as is. If needed, add more patterns to config.

Could be expanded to ticket creation if needed (although we have bulk resolve for that so not as critical).

#### Checklist
- [x] Included tests
- [x] Tested in [MCL-16427](https://bugs.mojang.com/browse/MCL-16427)

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -441,6 +441,26 @@
           "Works As Intended"
         ]
       },
+      "removeSpam": {
+        "resolutions": [
+          "Awaiting Response",
+          "Cannot Reproduce",
+          "Done",
+          "Duplicate",
+          "Fixed",
+          "Incomplete",
+          "Invalid",
+          "Unresolved",
+          "Won't Fix",
+          "Works As Intended"
+        ],
+        "patterns": [
+          {
+            "pattern": "\\[~",
+            "threshold": 5
+          }
+        ]
+      },
       "resolveTrash": {
         "excludedStatuses": ["Postponed"],
         "resolutions": [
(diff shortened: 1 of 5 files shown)


src/test/kotlin/io/github/mojira/arisa/modules/RemoveSpamModuleTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        val issue = mockIssue(
+            comments = listOf(comment)
+        )
+
+        val result = module(issue, YESTERDAY)
+
+        result.shouldBeLeft(OperationNotNeededModuleResponse)
+    }
+
+    "should hide comment when patter is triggered very often" {
```
SPGoding, line 84:
```suggestion
    "should hide comment when pattern is triggered very often" {
```

SPGoding (approved):

arisa.json
```diff
@@ -441,6 +441,26 @@
...
+          "Fixed",
+          "Incomplete",
+          "Invalid",
+          "Unresolved",
+          "Won't Fix",
+          "Works As Intended"
+        ],
+        "patterns": [
+          {
+            "pattern": "\\[~",
```
urielsalis, line 459: Can we make this check that we are closing the ] in the same word too?

Also I think this won't trigger if it doesn't start with a ping

urielsalis (approved):

arisa.json
violine1101: > Can we make this check that we are closing the ] in the same word too?

Yeah, technically that would be possible too but it makes the regex more error prone and complicated, while not exactly making much of a difference. I don't think there's any reason for the string `[~` being used for anything other than a user mention.

> Also I think this won't trigger if it doesn't start with a ping

I don't see why that would be the case. I've tested it, just to make sure, and it appears to work fine. [MCL-16427](https://bugs.mojang.com/browse/MCL-16427)

PR #539, opened 2021-02-19
Fix wrong e-mail regex of PrivacyModule

Marcono1234:
## Purpose
It appears the regex syntax has a peculiarity where `[…-]` matches the `-` literally (while `[…-…]` represents a range). When #403 added the `_` it accidentially converted `-` into ranges which match a bunch of characters which are not allowed by [RFC 952](https://tools.ietf.org/html/rfc952).

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
@@ -12,7 +12,7 @@ class PrivacyModule(
 ) : Module {
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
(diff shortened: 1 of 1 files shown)


SPGoding:
I'm pretty sure the escape here is unnecessary :eyes:

SPGoding (approved):

Marcono1234:
The last escape is not strictly needed (as pointed out in the description), but if we omit it we will likely run into the same problem again as #403 did when we add another character.

To verify that it makes a difference, run
```java
"/:;<=>?@[]\\^".matches("[a-zA-Z0-9.-_]+")
```
The result will be `true`.

I don't have the development setup for Arisa locally, but feel free to add tests for this to verify that it does not include any of these characters.

PR #544, opened 2021-02-20
Fix #537

### LINKED ISSUE #537: Remove reporter after continued violation of the RemoveVersion Module
## The Problem
Currently, users like to fight against the bot and try to add an affected version to their report multiple times, see https://bugs.mojang.com/browse/MC-215560, for an example.
## Possible Solution
After Arisa comments on the report with a warning a specific number of times (5), Arisa will then set the reporter of the ticket to NoUser

urielsalis:
## Purpose
Fix #537
## Approach
Detect from the changelog if arisa removed a version 5 or more times, and change the reporter to nouser
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
@@ -20,14 +20,14 @@ class RemoveVersionModule(
                 .map { it.remove }
             assertNotEmpty(removeAddedVersions).bind()
             removeAddedVersions.forEach(::run)
-            if (calledMoreThan5Times(changeLog)) {
+            if (calledMoreThanMaxTimes(changeLog)) {
                 changeReporter("NoUser")
             }
             addComment(CommentOptions(message))
         }
     }
 
-    private fun calledMoreThan5Times(changeLog: List<ChangeLogItem>): Boolean {
+    private fun calledMoreThanMaxTimes(changeLog: List<ChangeLogItem>): Boolean {
         return changeLog
             .asSequence()
             .filter { it.author.name == "arisabot" }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -17,10 +20,23 @@ class RemoveVersionModule(
...
             assertNotEmpty(removeAddedVersions).bind()
             removeAddedVersions.forEach(::run)
+            if (calledMoreThan5Times(changeLog)) {
+                changeReporter("NoUser")
+            }
             addComment(CommentOptions(message))
         }
     }
 
+    private fun calledMoreThan5Times(changeLog: List<ChangeLogItem>): Boolean {
```
chandler05, line 30: I suggest renaming the function to remove the "5", as we may want to change the value in the future

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -17,10 +20,23 @@ class RemoveVersionModule(
                 .map { it.remove }
             assertNotEmpty(removeAddedVersions).bind()
             removeAddedVersions.forEach(::run)
+            if (calledMoreThan5Times(changeLog)) {
```
SPGoding, line 23: This will be triggered on the sixth attempt; not sure if it's intended though.

SPGoding (approved):

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
urielsalis: the >= means it will match at 5 👁️

chandler05 (approved):

src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
SPGoding: But it's comparing with the amount of existing change log items, which doesn't include the current run (the sixth run).

urielsalis: ah, I dont think it matters that much in that case honestly, its just one extra time

PR #545, opened 2021-02-20
Fix #514

### LINKED ISSUE #514: Arisa not detecting certain transaction IDs
## The Bug
Arisa should be detecting all transaction IDs and making the ticket private.

## Example Ticket
https://bugs.mojang.com/browse/MC-212366

## Expected behavior
To make the ticket private

urielsalis:
## Purpose
Fix #514
## Approach
Add regexes from cube
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
@@ -13,7 +13,14 @@ class PrivacyModule(
 ) : Module {
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
-        """--accessToken ey""".toRegex()
+        """--accessToken ey""".toRegex(),
+        """[A-Z0-9]{17}""".toRegex(),
+        """([A-Z0-9]{4}-){3}[A-Z0-9]{4}""".toRegex(),
+        """braintree:[0-9]{6,7}""".toRegex(),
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -13,7 +13,14 @@ class PrivacyModule(
...
     private val patterns: List<Regex> = listOf(
         """\(Session ID is token:""".toRegex(),
-        """--accessToken ey""".toRegex()
+        """--accessToken ey""".toRegex(),
+        """[A-Z0-9]{17}""".toRegex(),
+        """([A-Z0-9]{4}-){3}[A-Z0-9]{4}""".toRegex(),
+        """braintree:[0-9]{6,7}""".toRegex(),
+        """([A-Z0-9]{5}-){5}[A-Z0-9]{5}""".toRegex(),
+        """([A-Za-z0-9]{4}-){3}[A-Za-z0-9]{4}""".toRegex(),
+        """[0-9]{3} [0-9]{3} [0-9]{4}""".toRegex(),
```
SPGoding, line 22: This one might also match coordinates?

urielsalis: good catch, should be rare but removed just in case

SPGoding (approved):

PR #547, opened 2021-02-20
Fix #493

### LINKED ISSUE #493: Add command to stop reopening of AR tickets unless original reporter responds.
## The Problem
Often A-R tickets are asking for specific information from the original reporter, such as exact messages they are seeing, steps to reproduce the issue, etc. Arisa reopens tickets when anyone responds, which is sometimes helpful, especially for bulk resolves, but for individual requests not so much.

## Possible Solution
Add ARISA_REOPEN_ANY to reopen if anyone responds. Default behavior would be to reopen only if the OP responds.

urielsalis:
## Purpose
Fix #493
## Approach
If ARISA_REOPEN_OP is present, only op is allowed to reopen
## Future work
I havent tested this in a real ticket, I should go to sleep
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/arisa.json
+++ b/arisa.json
@@ -251,19 +251,19 @@
           "Awaiting Response"
         ],
         "blacklistedRoles": [
-		  "staff",
-		  "global-moderators"
-		],
-		"blacklistedVisibilities": [
-		  "helper",
-		  "staff",
-		  "global-moderators"
-		],
-		"softARDays": 365,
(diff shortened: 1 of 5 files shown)


chandler05 (approved):

chandler05 (approved):

violine1101:
Added to the wiki

Issue #527, opened 2021-02-04
No (private) duplicates message is posted in case parent ticket is set to private


miwob:
Example ticket: https://bugs.mojang.com/browse/MC-213971

violine1101:
This is probably a more wide-spread issue, where if two modules try to do something, only one of them executes successfully. Also happens sometimes with CHK.

urielsalis:
MIght also cause it to leave empty comments #435

chandler05:
This happens whenever a future version is added on ticket creation on a ticket in a different language. Only one of the two happens.

PR #557, opened 2021-02-22
Improve config

violine1101:
## Purpose
Improves config in multiple areas:
* Use yml instead of JSON for easier configuration (especially when it comes to strings)
* Move config files into their own folder for better organization
* Use more intuitive config names defaults (e.g. replace `only` with `enabled`; `resolutions` is every resolution by default)

## Future work
I'm currently going through all modules and check what can be outsourced into the config. Also it may be possible to move most of the module registration to the modules themselves.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis:
> Also it may be possible to move most of the module registration to the modules themselves

I would love that, with an extra of in the future being able to separate modules into their own .jar (using a "library" for the config from the main arisa), meaning that for most changes we dont need a full redeploy and build of arisa (most changes are changing a module, config or adding a new module), we can turn off and on modules easily and the actual arisa code itself becomes way more testeable

.gitignore
```diff
@@ -2,11 +2,14 @@
...
 /lib/
 /out/
 /logs/
-last-run
\ No newline at end of file
+last-run
+
+# Old credentials config file
+arisa.yml
+/config/local.yml
```
urielsalis, line 15: I would put config/local.yml, so it works in windows

config/local.template.yml
```diff
@@ -0,0 +1,8 @@
+# Copy this file, rename it to `local.yml`, and enter your credentials to run the bot.
+arisa:
+  credentials:
+    username: <JIRA user name>
+    password: <plain text password>
+    dandelionToken: <Dandelion token>
```
urielsalis, line 6: This is no longer mandatory, so we can put a comment with optional

src/test/kotlin/io/github/mojira/arisa/ModuleRegistryTest.kt
```diff
@@ -1,57 +1,64 @@
...
-import io.kotest.matchers.shouldBe
+import io.kotest.matchers.collections.shouldNotContain
 import org.reflections.Reflections
 
+val CONFIG = Config { addSpec(Arisa) }
+    .from.yaml.watchFile("config/config.yml")
+    .from.map.flat(
+        mapOf(
+            // Overwrite these settings to ensure we're always covering both test cases
+            "arisa.modules.attachment.enabled" to "true",
```
urielsalis, line 18: Ideally we removed this, but well

urielsalis (commented):
Lets sync for the changes on the server side (copying it to both locations) so there is no downtime when moving

.gitignore
violine1101: I am on Windows and this is not an issue for me at least. `.gitignore` paths are relative to the directory in which the `.gitignore` is placed.

src/test/kotlin/io/github/mojira/arisa/ModuleRegistryTest.kt
violine1101: Yeah, I'm not quite happy either about having to use the real config file here and overwriting stuff but I couldn't quite figure out how to avoid this without specifying a completely separate test config file that would need to be updated along with the real one.

I guess this has the nice side effect of having a unit test that fails when the config isn't specified correctly?

violine1101:
I'll make separate PRs for the other things I had planned, they might take a bit longer to flesh out.

src/test/kotlin/io/github/mojira/arisa/ModuleRegistryTest.kt
violine1101: What exactly do you mean? I don't understand

config/local.template.yml
violine1101: Added a comment there, and added the other optional credentials too

src/test/kotlin/io/github/mojira/arisa/ModuleRegistryTest.kt
urielsalis: make sure that config.yml has arisa.modules.attachment.enabled as true and arisa.modules.missingCrash.enabled as false so we dont need to set it for the tests, we can disable in local.yml if we dont want it on production

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
```diff
@@ -131,9 +131,13 @@ class ModuleExecutor(
         onQueryNotAtResultEnd: () -> Unit,
         newPostedCommentCache: Cache<MutableSet<String>>
     ): List<Issue> {
-        val projects = (config[moduleConfig.projects] ?: config[Arisa.Issues.projects])
-        val resolutions = config[moduleConfig.resolutions].map(String::toLowerCase)
+        val projects = config[moduleConfig.projects] ?: config[Arisa.Issues.projects]
```
urielsalis, line 134: I really wish we could make this entire class more testable, but out of scope for this PR

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
```diff
@@ -59,15 +59,14 @@ class ModuleRegistry(private val config: Config) {
...
-    fun getModules(): List<Entry> {
-        val onlyModules = modules
-            .filter { config[it.config.only] }
-        return if (onlyModules.isEmpty()) {
-            modules
-        } else {
-            onlyModules
-        }
-    }
+    fun getAllModules(): List<Entry> = modules
```
urielsalis, line 62: I really wish we could make this entire class more testable, but out of scope for this PR

urielsalis (approved):

SPGoding (approved):

README.md
```diff
@@ -45,12 +45,16 @@ git clone https://github.com/mojira/arisa-kt.git
...
-To run the bot, you need to run the following command and it will use the default configuration:
+To run the bot, you need to run the following command, and it will use the default configuration:
 ```
 ./gradlew build installDist
 ./build/install/arisa-kt/bin/arisa-kt
 ```
 
+In order to configure Arisa, create a `local.yml` file inside the `config` folder.
+There you can set all your credentials and overwrite any configs in `config.yml` for debugging purposes.
+There are also some specific debug configs, you can find their documentation in the `ArisaConfig.kt` file.
```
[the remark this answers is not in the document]
violine1101: I figured just mentioning it there in the readme should be enough, this is really only relevant if you are working on Arisa

src/main/kotlin/io/github/mojira/arisa/ModuleRegistry.kt
violine1101: Yeah :/ I needed to add that method just for the testing. Would have liked to avoid it.

src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt
violine1101: This class probably needs to be rewritten anyway because of #527

Issue #553, opened 2021-02-21
Improve logging and error handling


violine1101:
## The Problem
Currently, logging is very vague, often just specifying whether or not a module fails or succes succeeds. This makes debugging very hard when something fails.

## Possible Solution
Add logging messages to `ModuleResponse` and log them after the module has been executed

violine1101:
oops that was wrong

PR #560, opened 2021-02-22
Fix logging

violine1101:
## Purpose
Currently the logging files are saved in the root directory instead of a separate folder

## Approach
Changed the logging folder and logging file names.

Additional changes:
- Use plain text instead of HTML
- Save 14 days of logs
- Remove unneeded info from log (thread, logger)
  - We're currently only logging on a single thread and a single logger, so we don't need to log which thread and logger we're using
  - I plan to get rid of the separate `LanguageModule` logger along with #553. For now it should be obvious what the log message is from.

## Future work
Make error logging actually useful (#553)

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -1,25 +1,25 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %-5level - %msg%n</pattern>
         </encoder>
     </appender>
 
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>logs\arisakt.html</file>
+        <file>logs/arisakt.log</file>
         <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
             <level>INFO</level>
         </filter>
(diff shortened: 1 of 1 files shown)


src/main/resources/logback.xml
```diff
@@ -1,25 +1,25 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %-5level - %msg%n</pattern>
```
urielsalis, line 4: I would keep logger, as some modules have their own logger

violine1101: The only module that uses its own logger currently is `LanguageModule`. Do you think it makes sense to add loggers to other modules too? I'd have gone the route of modifying the ModuleResponse to return arbitrary messages from the modules instead of having an extra logger for each module

urielsalis: I think it's cleaner if modules log themselves instead of sending that over somewhere

violine1101: Hmm, yeah after thinking about it it's probably best to keep thread and logger there, just in case for future expansion, and it also doesn't really hurt to have them there.

I have an idea on how to improve module feedback and logging, but I'll first need to experiment a bit to see if it works

urielsalis (approved):

PR #562, opened 2021-02-23
Fix sentences with the first letter of each word capitalized

### LINKED ISSUE #424: Fix Sentences That Have The First Letter Capitalized For Every Word
## The Problem

Some Users Capitalize Every First Letter Of The Words In Their Bug Description, Which Makes It Fairly Difficult To Read. It Would Be Great If Arisa Could Edit Them To The Human-Friendly Format.

## Example Ticket

- [Https://Bugs.Mojang.Com/Browse/Mce-2612](https://bugs.mojang.com/browse/MCE-2612)

## Possible Solution

Use A Regular Expression To Match These Sentences To Decide If The Module Should Execute? I'm Not Actually Sure.

## Concerns

Some Letters That Should Actually Be Capitalized Might Be Lower-Cased By The Bot. It Shouldn'T Be A Big Deal Though. We Could Do A Special Case For The Single Letter "`I`".

chandler05:
## Purpose
Fix #424 
## Future work
I didn't feel like adding any special cases, and its not that big of a deal. We may add them in the future, but idk
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/test/kotlin/io/github/mojira/arisa/modules/FixCapitalizationModuleTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/modules/FixCapitalizationModuleTest.kt
@@ -1,0 +1,50 @@
+package io.github.mojira.arisa.modules
+
+import arrow.core.right
+import io.github.mojira.arisa.utils.RIGHT_NOW
+import io.github.mojira.arisa.utils.mockIssue
+import io.kotest.assertions.arrow.either.shouldBeLeft
+import io.kotest.assertions.arrow.either.shouldBeRight
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+
+private val A_SECOND_AGO = RIGHT_NOW.minusSeconds(1)
+
+class FixCapitalizationModuleTest : StringSpec({
+    val module = FixCapitalizationModule()
+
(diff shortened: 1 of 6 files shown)


src/test/kotlin/io/github/mojira/arisa/modules/FixCapitalizationModuleTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val result = module(issue, A_SECOND_AGO)
+
+        result.shouldBeLeft(OperationNotNeededModuleResponse)
+    }
+
+    "should replace capitalized sentences in description" {
+        var hasUpdatedDescription: String? = null
+
+        val issue = mockIssue(
+            description = "Testing Except Capitalized.",
```
urielsalis, line 38: I would make it run only if every single word that starts with a letter and it's not a link is capitalized, else it would be annoying

chandler05: That's what it does, I think?

chandler05: Yeah, it only edits the capitalization if every single word in the sentence starts with a capital letter

PR #567, opened 2021-02-24
Fix NPE in getForwardResolvedOrNot

SPGoding:
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


chandler05 (approved):

Marcono1234:
I am not really familiar with Kotlin, but would it be possible to write these extension functions (there appear to be more than just `getCommonFieldOrNull`)  in a way that declares the return type?
Then the return type would have to be nullable and the compiler would have caught this error, right?

urielsalis:
> I am not really familiar with Kotlin, but would it be possible to write these extension functions (there appear to be more than just getCommonFieldOrNull) in a way that declares the return type?

Yes, you can  put fun name(arg: Arg): ReturnType

Add a ? for nullable

Anything coming from Jira can be null, as its a java library, and we need to check it 👁️

PR #569, opened 2021-02-27
Small fixes for capitalization regex

chandler05:
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixCapitalizationCommand.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixCapitalizationCommand.kt
@@ -3,6 +3,7 @@ package io.github.mojira.arisa.modules.commands
 import io.github.mojira.arisa.domain.Issue
 
 class FixCapitalizationCommand : Command1<String> {
+    @Suppress("MaxLineLength")
     override operator fun invoke(issue: Issue, arg: String): Int {
         val capitalizationRegex =
                 """(?<=\.\s|^|!\s|\?\s|\n)[A-Z][A-Za-z-'0-9]*((\s|,\s|;\s|:\s)[A-Z][A-Za-z-'0-9]*)*(?=\.|${'$'}|!|\?|\n)"""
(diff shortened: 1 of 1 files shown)


urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/commands/FixCapitalizationCommand.kt
```diff
@@ -3,9 +3,10 @@ package io.github.mojira.arisa.modules.commands
 import io.github.mojira.arisa.domain.Issue
 
 class FixCapitalizationCommand : Command1<String> {
+    @Suppress("MaxLineLength")
     override operator fun invoke(issue: Issue, arg: String): Int {
         val capitalizationRegex =
-                """(?<=\.\s|^|!\s|\?\s)[A-Z][A-Za-z0-9]*((\s|,\s|;\s|:\s)[A-Z][A-Za-z-'0-9]*)*(?=\.|${'$'}|!|\?)"""
+                """(?<=\.\s|^|!\s|\?\s|\n)[A-Z][A-Za-z-'0-9]*((\s|,\s|;\s|:\s)[A-Z][A-Za-z-'0-9]*)*(?=\.|${'$'}|!|\?|\n)"""
```
Marcono1234, line 9: Could you please escape the `-`? Otherwise we might run into the same issue as #539 when someone adds another character in front of `-'` turning it into a range.
```suggestion
                """(?<=\.\s|^|!\s|\?\s|\n)[A-Z][A-Za-z\-'0-9]*((\s|,\s|;\s|:\s)[A-Z][A-Za-z\-'0-9]*)*(?=\.|${'$'}|!|\?|\n)"""
```

Additionally it might be good to add some comments explaining what this should match. For example I don't quite understand the intention behind these:
- `[A-Z][A-Za-z-'0-9]`: Why match one (or more) letters followed by a number?, e.g. `Test1`
Sure you can lowercase this as well, but is such a string really common?
- `${'$'}`: Why match _end of line_ `{'` _end of line_ `'}`

chandler05: > ${'$'}: Why match "end of line {' end of line '}"?

Oops, that was IntelliJ's fault.
IntelliJ formats it as such whenever you copy and paste in the regex

> [A-Z][A-Za-z-'0-9]: Why match one (or more) letters followed by a number?, e.g. Test1
Sure you can lowercase this as well, but is such a string really common?

I put that in there to continue matching when users have the version number in the sentence (1.16.5, 21w08b etc.) but on second that, it won't even match that 😅

chandler05: I don't see any reason to remove the numbers though

PR #573, opened 2021-03-01
Stop seeds from being privated

### LINKED ISSUE #571: Arisa detects seeds as transaction IDs
## The Bug
Arisa detects any string of 17 numbers as a transaction ID. However, these are far more likely seeds.

## Example Ticket
https://bugs.mojang.com/browse/MC-217404

## Expected behavior
Don't detect those types of transaction IDs.

The current regexes are far too error-prone, although I don't really know how that could be changed easily.

chandler05:
## Purpose
Fix #571 
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
@@ -14,7 +14,7 @@ class PrivacyModule(
     private val patterns: List<Regex> = listOf(
         """.*\(Session ID is token:.*""".toRegex(),
         """.*--accessToken ey.*""".toRegex(),
-        """.*\b[A-Z0-9]{17}\b.*""".toRegex(),
+        """.*\b(?=[^\s]*[A-Z])(?=[^\s]*[0-9])[A-Z0-9]{17}\b.*""".toRegex(),
         """.*\b([A-Z0-9]{4}-){3}[A-Z0-9]{4}\b.*""".toRegex(),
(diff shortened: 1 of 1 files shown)


urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -14,7 +14,7 @@ class PrivacyModule(
     private val patterns: List<Regex> = listOf(
         """.*\(Session ID is token:.*""".toRegex(),
         """.*--accessToken ey.*""".toRegex(),
-        """.*\b[A-Z0-9]{17}\b.*""".toRegex(),
+        """.*\b(?=[^\s]*[A-Z])(?=[^\s]*[0-9])[A-Z0-9]{17}\b.*""".toRegex(),
```
Marcono1234, line 17: This mismatch between `\b` before and behind the regex but `\s` in the look aheads now causes a different issue where any substring consisting only of letters or only of digits matches when followed by `\b` chars which are not in `\s`. E.g. the following will now be matched:
```
AAAAAAAAAAAAAAAAA;;;;;;;;;;;;;;;;;;;;;;;;;;;;;1
```
To solve this either use the exact regex (with leading and trailing `.*`) as described in your comment https://github.com/mojira/arisa-kt/issues/571#issuecomment-788057416, or use the alternative with `\b` as described in https://github.com/mojira/arisa-kt/issues/571#issuecomment-788070724.

PR #579, opened 2021-03-04
Refactor RemoveUserCommand

### LINKED ISSUE #515: Cleanup code from #512
## The Problem
#512  was merged quickly to get rid of spammers, code should be cleaned up and tested

violine1101:
## Purpose
When using the RemoveUserCommand with some user names, nothing happens because the user name is not recognized correctly. This PR aims to fix that.

Closes #515.

## Approach
Stop using an RSS feed to get user activity; use a filter instead.

On the way to this, we need to refactor a lot because the current logic for querying something from a filter is concentrated (and scattered around) in the main loop of the bot.

I also added the following changes:
* Change PurgeAttachmentCommand to take a username and an optional min attachment ID (**TODO: There should probably be an optional max ID arg too**)
* Started to add tests for RemoveUserCommand
* Add ListUserActivityCommand for debugging purposes
* Replace all the comment caches that are being passed around by a singleton `CommentCache` (**TODO: Test that!**)

## Future work
Refactor more.

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -35,6 +36,46 @@ fun connectToJira(username: String, password: String, url: String): Pair<JiraCli
     return JiraClient(url, credentials) to credentials
 }
 
+/**
+ * Returns a list of strings indicating ticket IDs that are contained in the given jql filter.
+ * TODO: Actually return the tickets themselves instead of only ticket IDs.
+ *
+ * @param cap - How many tickets should be queried at most. No cap by default.
+ */
+fun searchIssues(jiraClient: JiraClient, jql: String, cap: Int = Int.MAX_VALUE) = runBlocking {
```
urielsalis, line 45: You are not using this cap you are still using MAX_RESULTS 👀

src/test/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommandTest.kt
```diff
@@ -0,0 +1,54 @@
...
+            { _, _ ->
+                Either.fx {
+                    calledSearch = true
+                    issueList
+                }
+            },
+            { ticketId ->
+                Either.fx {
+                    queriedIssues.add(ticketId)
+                    throw RuntimeException("Cannot mock a rcarz issue")
```
urielsalis, line 32: We cannot? mock(JiraIssue.class) doesnt work?

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
violine1101: I'm using the cap in L68

urielsalis: you can put it instead of MAX_RESULTS and then we dont query as much data from Jira :D (I think its set to 200), 

That also means you dont need to check on each loop

src/test/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommandTest.kt
violine1101: I couldn't figure out how it would work. And I don't know what `mock(JiraIssue.class)` would be? `JiraIssue` isn't a class and `mock` isn't a function that I know of.

All the fields in the rcarz issue are private so I wouldn't know how to mock them and I don't want to waste time writing a mock for a library that we won't use in the future anyway. And it'd be good to refactor this anyway so that I don't even need to worry about rcarz issues.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
violine1101: Yep, true. I added the cap after the fact just as a safety measure. Would be better to move it directly to replace `MAX_RESULTS` -- then we can also get rid of the loop altogether.

src/test/kotlin/io/github/mojira/arisa/modules/commands/RemoveUserCommandTest.kt
violine1101: Aaah I think I've found what you mean. Brb doing that.

violine1101: Ok but this is still very annoying to mock even with mockK. I guess we're just gonna need to have this test incomplete for now.

violine1101:
The current testing for `RemoveUserCommand` is a mess but I didn't want to interfere with our plans to refactor issue. It works as it is currently, although it's not very nice code.

urielsalis (approved):

PR #581, opened 2021-03-05
Revert "Don't test draft pull requests"

SPGoding:
Reverts mojira/arisa-kt#256


chandler05 (approved):

chandler05:
Not going to merge until we discuss this, but speaking of which, why do we even need to discuss this?

SPGoding:
I'd suggest we discuss if we need to discuss this.

PR #582, opened 2021-03-06
Prevent Arisa from commenting the duplicate message with a comment

### LINKED ISSUE #492: Add a comment that stops arisa from posting the 'duplicate' message
## The Problem
In some situations you may not want the 'duplicate' message to be posted
## Possible Solution
Make a comment that helper+ can use to stop the message from being sent

chandler05:
## Purpose
Close #492
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


Greymagic27:
Wondering if the command could be ARISA_NO_DUPE_MESSAGE instead. Just a little bit less to write

urielsalis (approved):

PR #584, opened 2021-03-06
Major refactor of domain and infrastructure

### LINKED ISSUE #527: No (private) duplicates message is posted in case parent ticket is set to private

urielsalis:
## Purpose
Make jira completly independent from our logic and testable, and cleanup the code

Fixes #527
## Approach

- Create new domain classes that store all the state
- Modules see the modified state after other modules execute
- All changes per issue saved at once
- Proper hexagonal/onion/whatever arch (domain only sees domain, modules only see domain and themselves, infra sees everything)
- 
## Future work
- See help needed!

## Help needed
This is a work in progress! Feel free to jump in and help, mark things as done or add new ones!
Ordered by priority:
- [x] Create the domain classes with the state
- [x] Map from Jira
- [x] Map to Jira
- [x] Modify the module executor to keep the issues 
- [x] Remove everything that we can from ArisaMain into their own classes. ArisaMain should be as clean as possible (only create instances of services needed and do the main loop)
- [x] Re add the debug ticket whitelist
- [x] Re add running failed/rerun tickets
- [x] Re add filters for resolutions, status, etc
- [x] We need to prevent comments being added again the next run again
- [x] Cleanup helpers so we dont need to pass a issue every time (if possible)
- [x] Fix Helpers.kt links functions that prevent compile
- [x] Make sure we replace keys before adding the comment
- [ ] Each module has to be adapted :D 
- [ ] Make sure it compiles and passes ktlint/detekt
- [ ] Make sure existing tests pass / are removed if they are no longer needed
- [ ] Decide what to do if getting the message fails in Issue.kt for addMessage
- [x] Create the concept of a delayed module (runs on tickets older than 24 hs) and re enable duplicatedmessage and linked using that
- [ ] Error checking, im just letting exceptions bubble up but we should probably catch and handle it some of them
- [ ] I didnt touch the module executor or the other classes that much, im sure they could be cleaned up too
- [ ] Tests tests tests tests

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


violine1101:
> Create the concept of a delayed module (runs on tickets older than 24 hs) and re enable duplicatedmessage and linked using that

It's probably a good idea to not include that in this PR yet. IMO we should first focus on domain, then change the executor to no longer trigger a different search request for each module in a different refactor. If we do too many things at once, things are more likely to break, and it should be fairly easy to simply change out the domain stuff without interfering with anything else.

urielsalis:
For the single query thing, it's actually because it's way easier that way. Else we would have to sincronize the changes between modules and the issues instead of just sending the same list to all of them

urielsalis:
Error checking I mean in the registry and executors, as during the refactor I removed some checks

violine1101:
> For the single query thing, it's actually because it's way easier that way. Else we would have to sincronize the changes between modules and the issues instead of just sending the same list to all of them

Hmm true. I guess the easiest solution for that is then to group modules by jql query and just pass the queried tickets along the tickets in each group.

Although duplicate and linked modules have different JQLs, so they would need to each have their own group with no other modules in them.

urielsalis:
I wouldnt group all of them. Just execute modules with custom jql separately

PR #594, opened 2021-04-22
Rename 'ARISA_NO_DUPLICATE_MESSAGE' and 'ARISA_REMOVE_USER'

### LINKED ISSUE #588: Change new 'ARISA_NO_DUPLICATE_MESSAGE' to 'ARISA_NO_DUPE_MSG' or similar
This isn't anything major but I feel like ARISA_NO_DUPLICATE_MESSAGE is a mouthful to write. Maybe condensing it down might be an idea?

### LINKED ISSUE #590: Rename 'ARISA_REMOVE_USER'
## The Problem
Currently this command is misleading as rather than removing a user, it instead restricts their comments.
## Possible Solution
A solution would be to change it to ARISA_RESTRICT_COMMENTS or similar.

https://github.com/mojira/arisa-kt/wiki/Commands#arisa_remove_user

chandler05:
## Purpose
Close #588 and close #590
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -118,7 +118,11 @@ arisa:
         MC-128302: duplicate-of-mc-128302
         MCL-5638: duplicate-of-mcl-5638
       privateMessage: duplicate-private
-      preventMessageTag: ARISA_NO_DUPLICATE_MESSAGE
+      preventMessageTags:
+        - ARISA_NO_DUPLICATE_MESSAGE
+        - ARISA_NO_DUPE_MESSAGE
+        - ARISA_NO_DUPLICATE_MSG
+        - ARISA_NO_DUPE_MSG
       resolutionMessages:
         Fixed: duplicate-fixed
         Works As Intended: duplicate-wai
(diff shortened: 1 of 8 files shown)


SPGoding (approved):

config/config.yml
```diff
@@ -118,7 +118,11 @@ arisa:
         MC-128302: duplicate-of-mc-128302
         MCL-5638: duplicate-of-mcl-5638
       privateMessage: duplicate-private
-      preventMessageTag: ARISA_NO_DUPLICATE_MESSAGE
+      preventMessageTags:
+        - ARISA_NO_DUPLICATE_MESSAGE
+        - ARISA_NO_DUPE_MESSAGE
+        - ARISA_NO_DUPLICATE_MSG
+        - ARISA_NO_DUPE_MSG
```
Marcono1234, line 125: I am not sure if it is really a good idea to have multiple alternatives here.
Keeping `ARISA_NO_DUPLICATE_MESSAGE` for legacy reasons is fine, but I think only one additional short tag (e.g. `ARISA_NO_DUPE_MSG`) should be used then. Otherwise it might cause confusion, what the "right" tag is

PR #597, opened 2021-04-27
Reformat attachment details command to be more readable

chandler05:

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
@@ -29,13 +29,8 @@ class AttachmentModule(
     }
 
     private fun List<Attachment>.getCommentInfo() = this
-        .run {
-            when (size) {
-                1 -> "[~${get(0).uploader!!.name}]: ${get(0).name}"
-                else -> subList(0, lastIndex)
-                    .map { "[~${get(0).uploader!!.name}]: ${get(0).name}\n" }
-            }
-        }
+        .map { "- [~${it.uploader!!.name}]: ${it.name}" }
+            .joinToString { "\n" }
 
     private fun endsWithBlacklistedExtensions(extensionBlackList: List<String>, name: String) =
         extensionBlackList.any { name.endsWith(it) }
--- a/src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
@@ -37,7 +37,7 @@ class AttachmentModule(
                 2 -> "[${get(0)}|https://bugs.mojang.com/secure/ViewProfile.jspa?name=${get(0)}] " +
                         "and [${get(1)}|https://bugs.mojang.com/secure/ViewProfile.jspa?name=${get(1)}]"
                 else -> subList(0, lastIndex)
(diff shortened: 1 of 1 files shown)


Marcono1234:
Looks fine to me (though I am not very familiar with Kotlin). Though I am not sure if keeping the additional logic for building proper English sentences (instead of just joining the names with `, `)  is really worth it.

Additionally, if I recall correctly some usernames containing special chars caused issues in the past, though I am not sure whether that applied to https://bugs.mojang.com/secure/ViewProfile.jspa?name=.

Maybe it would be easier to use a username link <code>[~<i>user</i>]</code> (see [Jira help](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=links)) than to use a URL?

chandler05:
I temporarily forgot that existed. I'll add it instead of the links

Marcono1234:
Thanks!
I have edited my comment multiple times, could you please have a look at it again in case you have not seen the latest edits yet?

src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
```diff
@@ -19,33 +19,21 @@ class AttachmentModule(
...
             assertNotEmpty(functions).bind()
-            val usernames = functions.getUsernames()
-            val attachmentNames = functions.getAttachmentNames()
+            val commentInfo = functions.getCommentInfo()
             functions
                 .map { it.remove }
                 .forEach { it.invoke() }
             addComment(CommentOptions(attachmentRemovedMessage))
-            addRawRestrictedComment("Attachment Details:\nFilename: $attachmentNames\nUploader: $usernames", "helper")
+            addRawRestrictedComment("Attachments:\n$commentInfo", "helper")
```
Marcono1234, line 27: Might be good to make it a little bit clearer what this comment describes:
```suggestion
            addRawRestrictedComment("Removed attachments:\n$commentInfo", "helper")
```

src/main/kotlin/io/github/mojira/arisa/modules/AttachmentModule.kt
```diff
@@ -19,33 +19,21 @@ class AttachmentModule(
...
-        .run {
-            when (size) {
-                1 -> get(0)
-                2 -> "${get(0)} and ${get(1)}"
-                else -> "${subList(0, lastIndex).joinToString(", ")}, and ${last()}"
+                1 -> "[~${get(0).uploader!!.name}]: ${get(0).name}"
+                else -> subList(0, lastIndex)
+                    .map { "[~${get(0).uploader!!.name}]: ${get(0).name}\n" }
             }
         }
```
Marcono1234, line 38: It looks like currently the return type is `Any` because `map` returns a `List`. Additionally `subList(0, lastIndex)` would exclude the last element, wouldn't it?
What do you think about the following instead?
```kotlin
    private fun List<Attachment>.getCommentInfo() =
        map { " - [~${it.uploader!!.name}]: ${it.name}" }
            .joinToString { "\n" }
```

chandler05: Looks like a better solution 👍

Marcono1234 (approved):
Looks good to me, though feel free to wait for a review from someone with more Kotlin experience.

urielsalis (approved):

PR #602, opened 2021-05-03
Fix errors when trying to empty the affected versions field

### LINKED ISSUE #601: Arisa fails to revert a change to affected versions on a resolved report when only one version is affected and two versions are switched
## The Bug
When a resolved ticket has only one affected version, and the reporter switches the affected version for another version (in the same action), Arisa tries but fails to remove the new affected version. This is probably because it can't empty the field (it tries to).
## Example Ticket
MC-222695 (private), and https://bugs.mojang.com/browse/MC-224653 (public)
## Expected behavior
Arisa removes the new affected version and replaces it with the old one, or Arisa does nothing (no attempt to complete an action it can't complete)

chandler05:
## Purpose
Fix #601 
## Approach
Don't remove the final version with the RemoveVersion module
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
@@ -14,6 +14,7 @@ class RemoveVersionModule(
 ) : Module {
     override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
         Either.fx {
+            assertGreaterThan(affectedVersions.size, 1)
             val addedVersions = getExtraVersionsLatelyAddedByNonVolunteers(lastRun)
             val removeAddedVersions = affectedVersions
                 .filter { it.id in addedVersions }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/RemoveVersionModule.kt
```diff
@@ -14,6 +14,7 @@ class RemoveVersionModule(
 ) : Module {
     override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
         Either.fx {
+            assertGreaterThan(affectedVersions.size, 1).bind()
```
Marcono1234, line 17: Maybe it would be saner to verify that when the amount of removed versions is `< affectedVersions`, otherwise the same issue might occur again when two versions are added at the same time.

So assert that `removeAddedVersions.size < affectedVersions.size`.

chandler05: Good point, fixed

Marcono1234: Really minor, but might be better to compare with `removeAddedVersions`. Otherwise the module would refuse to remove versions when the versions are not part of the affected versions anymore (and would therefore not be part of `removeAddedVersions` either).

chandler05: Yep, and I think that was causing the error. Thanks

urielsalis (approved):

PR #609, opened 2021-05-07
Leave staff comment when staff tries to reopen keep private ticket 

chandler05:


chandler05:
Not sure why the test is failing :/

urielsalis:
I would make Arisa leave a staff comment telling you to remove the tag first, not ignore it

chandler05:
Ok its been changed to the comment, still have no idea why its failing. I'm probably just tired and can't see it...

src/test/kotlin/io/github/mojira/arisa/modules/KeepPrivateModuleTest.kt
```diff
@@ -90,6 +98,31 @@ class KeepPrivateModuleTest : StringSpec({
...
+            changeLog = listOf(REMOVE_SECURITY_STAFF),
+            setPrivate = { didSetToPrivate = true; Unit.right() },
+            addRawRestrictedComment = { _, _ -> didComment = true; Unit.right() }
+        )
+
+        val result = module(issue, RIGHT_NOW)
+
+        result.shouldBeRight(ModuleResponse)
+        didSetToPrivate shouldBe true
+        didComment shouldBe true
```
urielsalis, line 123: This is failing(returning false)

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -24,7 +24,15 @@ class KeepPrivateModule(
 
             val markedTime = comments.first(::isKeepPrivateTag).created
             val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
-            if (changedTime != null && changedTime.isAfter(markedTime)) {
+            val securityChange = changeLog
+                .lastOrNull() { isSecurityChangeToPublic(it, securityLevel, project.privateSecurity) }
+            if (
+                securityChange != null && securityChange.getAuthorGroups()
+                    ?.any { it == "global-moderators" || it == "staff" } == true
```
Marcono1234, line 31: Can the `== true` be omitted here?

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -24,7 +24,15 @@ class KeepPrivateModule(
 
             val markedTime = comments.first(::isKeepPrivateTag).created
             val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
-            if (changedTime != null && changedTime.isAfter(markedTime)) {
+            val securityChange = changeLog
+                .lastOrNull() { isSecurityChangeToPublic(it, securityLevel, project.privateSecurity) }
+            if (
```
Marcono1234, line 29: Shouldn't `isSecurityChangeToPublic` apply to staff and non-staff actions, if so, could possibly combine this with calculating `val changedTime`.
Also, shouldn't for staff action the requirement `changedTime != null && changedTime.isAfter(markedTime)` apply as well?

Marcono1234:
> I would make Arisa leave a staff comment telling you to remove the tag first, not ignore it

This might not be effective when the staff member is not watching the report, but there is probably not a better solution to this.

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
chandler05: Still didn't solve the build failing though, and I have no idea why it's failing either.

Marcono1234: Yeah, my comment was a general recommendation and not related to the test failure.
Though maybe it would be good anyways to remove the `isSecurityChange` check and only use `securityChange` to determine the `changedTime`. Otherwise there are two different calls which determine the last relevant changelog entry, which might be error-prone and could cause issues in the future when they (for some reason) select different changelog entries.

src/test/kotlin/io/github/mojira/arisa/modules/KeepPrivateModuleTest.kt
chandler05: Marcono figured it out 👍

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -23,9 +23,21 @@ class KeepPrivateModule(
...
+                .lastOrNull() { isSecurityChangeToPublic(it, project.privateSecurity) }
+            val changedTime = securityChange?.created
             if (changedTime != null && changedTime.isAfter(markedTime)) {
-                addComment(CommentOptions(message))
+                if (
+                    securityChange.getAuthorGroups()
+                        ?.any { it == "global-moderators" || it == "staff" } == true
+                ) {
+                    addRawRestrictedComment(
+                        "To remove the security level," +
```
urielsalis, line 35: I would move this to helper-messages 👀

SPGoding (approved):

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -23,9 +23,21 @@ class KeepPrivateModule(
             setPrivate()
 
             val markedTime = comments.first(::isKeepPrivateTag).created
-            val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
+            val securityChange = changeLog
+                .lastOrNull() { isSecurityChangeToPublic(it, project.privateSecurity) }
+            val changedTime = securityChange?.created
             if (changedTime != null && changedTime.isAfter(markedTime)) {
-                addComment(CommentOptions(message))
+                if (
```
SPGoding, line 30: Let's open a "thread" here so the discussions don't get messy

PR #611, opened 2021-05-08
Fix and improve KeepPrivateModuleTest

Marcono1234:
## Purpose
Improvements and test fixes for #609

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
@@ -23,12 +23,12 @@ class KeepPrivateModule(
             setPrivate()
 
             val markedTime = comments.first(::isKeepPrivateTag).created
-            val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
             val securityChange = changeLog
-                .lastOrNull() { isSecurityChangeToPublic(it, securityLevel, project.privateSecurity) }
+                .lastOrNull() { isSecurityChangeToPublic(it, project.privateSecurity) }
+            val changedTime = securityChange?.created
             if (changedTime != null && changedTime.isAfter(markedTime)) {
                 if (
-                    securityChange != null && securityChange.getAuthorGroups()
+                    securityChange.getAuthorGroups()
                         ?.any { it == "global-moderators" || it == "staff" } == true
                 ) {
                     addRawRestrictedComment(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -23,12 +23,12 @@ class KeepPrivateModule(
             setPrivate()
 
             val markedTime = comments.first(::isKeepPrivateTag).created
-            val changedTime = changeLog.lastOrNull(::isSecurityChange)?.created
             val securityChange = changeLog
-                .lastOrNull() { isSecurityChangeToPublic(it, securityLevel, project.privateSecurity) }
+                .lastOrNull() { isSecurityChangeToPublic(it, project.privateSecurity) }
+            val changedTime = securityChange?.created
```
Marcono1234, line 28: As proposed in #609, uses `securityChange` to determine `changedTime` for consistency.

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
```diff
@@ -46,11 +46,8 @@ class KeepPrivateModule(
             comment.visibilityValue == "staff" &&
             (comment.body?.contains(keepPrivateTag!!) ?: false)
 
-    private fun isSecurityChange(item: ChangeLogItem) = item.field == "security"
-
-    private fun isSecurityChangeToPublic(item: ChangeLogItem, securityLevel: String?, privateLevel: String) =
-        item.field == "security" && item.changedFromString == privateLevel && item.changedToString == securityLevel &&
-                securityLevel != privateLevel
+    private fun isSecurityChangeToPublic(item: ChangeLogItem, privateLevel: String) =
+        item.field == "security" && item.changedFromString == privateLevel
```
Marcono1234, line 50: Removed two checks:
- `securityLevel != privateLevel`
That is probably redundant; module gets the last changelog entry matching, and also asserts that security level is not public, so (unless Jira is bugged?) `securityLevel != privateLevel` should always be true
- `item.changedFromString == privateLevel`
That is too restrictive; consider the hypothetical case where the security is changed from `private` -> `public` -> `other-public`. Then the module would not do anything because there exists no changelog entry from `private` -> `other-public` (also added a test for this)

src/test/kotlin/io/github/mojira/arisa/modules/KeepPrivateModuleTest.kt
```diff
@@ -13,14 +14,24 @@ import io.kotest.matchers.shouldBe
 private val REMOVE_SECURITY = mockChangeLogItem(
     created = RIGHT_NOW.minusSeconds(10),
     field = "security",
-    changedFromString = "10318",
```
Marcono1234, line 16: Tests were failing because they are using `"private"` as private security level of the mocked issue while these changelog items used `"10318"` (so tests were flawed previously as well); have now added a constant `PRIVATE_SECURITY_LEVEL` holding that security level.

Marcono1234 (commented):
@chandler05

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
chandler05: How about adding a check for `item.changedToString != privateLevel`? Just in case something happens and the security level changes from private to private.

src/test/kotlin/io/github/mojira/arisa/modules/KeepPrivateModuleTest.kt
chandler05: Ahhhh, thanks for finding this. I completely missed it myself.

src/main/kotlin/io/github/mojira/arisa/modules/KeepPrivateModule.kt
chandler05: Oh wait nevermind, that's also redundant.

chandler05 (approved):

Issue #281, opened 2020-05-23
Copy crash report to the description when it is attached


misode:
## The Problem
It is often useful to search for stack traces from crashes. Since Jira does not search through attachments, putting the head of the stack trace in the description is a good idea. Arisa could do this automatically when a crash report is attached.

## Possible Implementation
When a crash report is attached to a report for the first time, Arisa should check if the crash report isn't already present in the description. Detecting the Java exception name in a code block could be a good start. The crash report would be added at the bottom of the description

The crash report should be formatted as follows:
```
{code:title=(VERSION) [^CRASH_REPORT_NAME]}
Description: XYZ
Exception: XYZ
  at ...
  at ...
{code}
```
Example: https://bugs.mojang.com/browse/MC-181313

Marcono1234:
> if the crash report isn't already present in the description

Probably rather "if **a** crash report", otherwise it might add crash reports for multiple versions (unless that is desired).

urielsalis:
mc-crash-lib returns all the info to the crash module, so should be easy :D

urielsalis:
the exception field contains only the stacktrace

violine1101:
Personally I'd say it should only copy the stack trace initially after the bug report is first created. AKA only the reporter's crash report should be included. Afterwards we can take care of it manually.

urielsalis:
And if its a comment I dont mind if new ones are added, we can always just restrict it and a lot of the crash reports are added after the fact

PR #618, opened 2021-05-11
Add deobfuscated crashes

### LINKED ISSUE #606: Deobfuscate stack traces / crash reports
## The Problem
The fact that classes in the Minecraft JAR file are obfuscated has the following disadvantages:
- When a reporter just attaches a crash report without any additional information (possibly because they don't know when / why the game crashed), it is difficult to reason about what the cause might be
- Due to obfuscation mappings and line numbers changing between versions it can be difficult to tell whether two crash reports for different versions are caused by the same bug

It would therefore be useful if Arisa could deobfuscate stack traces / crash reports using the mapping files provided by Mojang.

## Possible Implementation
Initially it might be easiest to apply deobfuscation only to attached crash report files; detecting crash reports (or stack traces) in the description and comments might be difficult (and also Minecraft version information might be missing).

For deobfuscation ProGuard's [ReTrace](https://github.com/Guardsquare/proguard/blob/master/retrace/src/proguard/retrace/ReTrace.java) could be used.

Then, when either a new attachment is attached, or a report with attachments is created:
1. Check if the attachment is a crash report
2. Try to extract version information from the crash report (relying on Affected Versions of the issue might be error-prone)
3. Verify that the version has obfuscation mappings; i.e. version >= 1.14.4 or 19w36a
4. Check if the mapping file is already present on the file system
    1. If not, download the mapping file (perform sanity checks to prevent writing the mapping file outside of intented directory)
5. Deobfuscate the crash report
6. Attach the crash report as <code>(deobfuscated) <i>original-crash-report-name.txt</i></code>
(Maybe perform sanity check to prevent overwriting or shadowing existing attachment with that name)

## Considerations
- Verify that this use case is allowed by the license, in doubt get official confirmation from Mojang developers
- The obfuscation mapping license changed a few times (see also [Minecraft Wiki](https://minecraft.fandom.com/wiki/Obfuscation_map#License)), maybe deobfuscation for certain versions needs to be disabled if the license does not permit deobfuscation
- Check whether ReTrace can directly be applied to a crash report, or whether stack traces have to be extracted manually before being passed to ReTrace
- Check whether ["verbose" mode](https://github.com/Guardsquare/proguard/blob/master/docs/md/manual/tools/retrace.md#restoring-a-stack-trace-with-line-numbers-verbose--withverbose) of ReTrace might be useful (mainly for inspecting respective decompiled Minecraft source code more easily)
- ReTrace detects obfuscated class names using regex, this might lead to false positives when applying it to a complete crash report (since it contains more than just stack traces); though maybe this is negligible
- Maybe integrate this as part of https://github.com/urielsalis/mc-crash-lib ? Though arisa-kt would still be responsible for storing and providing mapping files then, and since most of the work is probably performed by ReTrace it might be easiest to have all code in arisa-kt only

urielsalis:
## Purpose
Fix #606
## Approach
Use newly added deobfuscation in mc-crash-lib and upload a new attachment with the suffix -deobfuscated.log to any new attachment + description
#### Checklist
- [/] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
@@ -12,6 +12,7 @@ import io.github.mojira.arisa.infrastructure.AttachmentUtils
 import io.github.mojira.arisa.infrastructure.config.CrashDupeConfig
 import me.urielsalis.mccrashlib.Crash
 import me.urielsalis.mccrashlib.CrashReader
+import java.io.File
 import java.time.Instant
 
 class CrashModule(
@@ -29,7 +30,9 @@ class CrashModule(
             val crashes = AttachmentUtils(crashReportExtensions, crashReader).extractCrashesFromAttachments(issue)
 
             assertContainsNewCrash(crashes, lastRun).bind()
+            uploadDeobfuscatedCrashes(issue, crashes)
             assertNoValidCrash(crashes).bind()
+
 
             val key = crashes
                 .sortedByDescending { it.first.created }
@@ -46,6 +49,20 @@ class CrashModule(
             }
         }
     }
+
+    private fun uploadDeobfuscatedCrashes(issue: Issue, crashes: List<Pair<AttachmentUtils.TextDocument, Crash>>) {
+        val minecraftCrashesWithDeobf = crashes
+            .map { it.first.name to it.second }
+            .filter { it.second is Crash.Minecraft }
+            .map { it.first to (it.second as Crash.Minecraft).deobf }
+            .filter { it.second != null }
+            .filterNot { issue.attachments.any { attachment -> attachment.name == getDeobfName(it.first) } }
+        minecraftCrashesWithDeobf.forEach {
+            issue.addAttachment(File(getDeobfName(it.first)))
+        }
+    }
+
+    private fun getDeobfName(name: String): String = "${name.substringAfterLast(".")}-deobfuscated.log"
 
     /**
      * Checks whether an analyzed crash report matches any of the specified known crash issues.
--- a/src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
@@ -173,6 +173,69 @@ const val JAVA_CRASH = """#
 # See problematic frame for where to report the bug.
 #"""
 
+const val OBFUSCATED_CRASH = """---- Minecraft Crash Report ----
+// Don't do that.
+
+Time: 30/04/21 17:23
+Description: mouseClicked event handler
+
+java.lang.OutOfMemoryError: Java heap space
+	at java.util.ArrayList.<init>(ArrayList.java:152)
+	at com.google.common.collect.Lists.newArrayListWithCapacity(Lists.java:190)
+	at mj$1.a(SourceFile:47)
+	at mj$1.b(SourceFile:32)
+	at md.b(SourceFile:471)
+	at md.a(SourceFile:32)
+	at md$1.a(SourceFile:83)
+	at md$1.b(SourceFile:69)
+	at md.b(SourceFile:471)
+	at md.a(SourceFile:32)
+	at md$1.a(SourceFile:83)
+	at md$1.b(SourceFile:69)
+	at md.b(SourceFile:471)
+	at md.a(SourceFile:32)
+	at md$1.a(SourceFile:83)
+	at md$1.b(SourceFile:69)
+	at mn.a(SourceFile:108)
+	at mn.a(SourceFile:75)
+	at mn.a(SourceFile:32)
+	at mn.a(SourceFile:26)
+	at cyg.a(SourceFile:229)
+	at cygLambda$2987/857564250.apply(Unknown Source)
+	at cyg.a(SourceFile:178)
+	at cyg.b(SourceFile:157)
+	at dsm.a(SourceFile:91)
+	at dsm.<init>(SourceFile:83)
+	at dsj.b(SourceFile:48)
+	at dot.b(SourceFile:325)
+	at djz.a(SourceFile:922)
+	at doy.d(SourceFile:141)
+	at doyLambda$2670/1502984812.onPress(Unknown Source)
+	at dlj.b(SourceFile:33)
+
+-- System Details --
+Details:
+	Minecraft Version: 1.16.5
+	Minecraft Version ID: 1.16.5
+	Operating System: Windows 10 (amd64) version 10.0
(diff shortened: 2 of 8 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -29,6 +30,7 @@ class CrashModule(
             val crashes = AttachmentUtils(crashReportExtensions, crashReader).extractCrashesFromAttachments(issue)
 
             assertContainsNewCrash(crashes, lastRun).bind()
+            uploadDeobfuscatedCrashes(issue, crashes)
```
violine1101, line 33: Shouldn't it just take the new crashes here? Not all of them? Otherwise it could upload attachments without end.

violine1101: Ah, no it doesn't do that because of Line 58, I see. Still it'd probably be better if it'd continue to work on the new crashes here instead of all of them, at least to save some performance.

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -47,6 +49,20 @@ class CrashModule(
...
 
+    private fun uploadDeobfuscatedCrashes(issue: Issue, crashes: List<Pair<AttachmentUtils.TextDocument, Crash>>) {
+        val minecraftCrashesWithDeobf = crashes
+            .map { it.first.name to it.second }
+            .filter { it.second is Crash.Minecraft }
+            .map { it.first to (it.second as Crash.Minecraft).deobf }
+            .filter { it.second != null }
+            .filterNot { issue.attachments.any { attachment -> attachment.name == getDeobfName(it.first) } }
+        minecraftCrashesWithDeobf.forEach {
+            issue.addAttachment(File(getDeobfName(it.first)))
```
violine1101, line 60: Where is the crash report file? Just in the root directory of Arisa? Or is it in some subfolder? Does it get deleted after it's been uploaded?

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -47,6 +49,20 @@ class CrashModule(
...
+        }
+    }
+
+    private fun getDeobfName(name: String): String = "${name.substringAfterLast(".")}-deobfuscated.log"
```
violine1101, line 64: Crash reports typically have file extension `.txt`, not `.log`.

violine1101: Oh also shouldn't it be `substringBeforeLast`, since you're stripping away the extension? Otherwise this would always be `txt-deobfuscated.log`, wouldn't it?

src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
```diff
@@ -173,6 +173,69 @@ const val JAVA_CRASH = """#
 # See problematic frame for where to report the bug.
 #"""
 
+const val OBFUSCATED_CRASH = """---- Minecraft Crash Report ----
```
violine1101, line 176: Why do we need this and can't just reuse `EXAMPLE_CRASH` or something similar?

Also this test file in general is massive and from what I can tell also partially tests mc-crash-lib. Instead of putting the entire crash report here, we should probably just mock the crash result or something.

It's ok for this PR but test organization is something that I'd like to look at in the future, atm it's very unweildy to navigate the test files and that's not good.

src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
```diff
@@ -707,20 +789,54 @@ class CrashModuleTest : StringSpec({
...
+            confirmationStatus = Unconfirmed,
+            priority = NoPriority,
+            resolveAsDuplicate = { resolvedAsDupe = true },
+            resolveAsInvalid = { resolvedAsInvalid = true },
+            addAttachment = { addedAttachement = true }
+        )
+
+        val result = module(issue, A_SECOND_AGO)
+
+        result.shouldBeLeft(OperationNotNeededModuleResponse)
```
violine1101, line 830: The `OperationNotNeededModuleResponse` irks me a bit here... but out of scope for this PR

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -12,6 +12,7 @@ import io.github.mojira.arisa.infrastructure.AttachmentUtils
 import io.github.mojira.arisa.infrastructure.config.CrashDupeConfig
 import me.urielsalis.mccrashlib.Crash
 import me.urielsalis.mccrashlib.CrashReader
+import java.io.File
 import java.time.Instant
 
 class CrashModule(
```
violine1101, line 18: It might make sense to split this module up at some point, it does so many things at once. (But out of scope for this PR)

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
urielsalis: Yep, gets deleted in the context. We cant use Temp files from java as they are only deleted on shutdown

src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
urielsalis: Tried reusing the other ones but none of them have the neccesary info (actually obfuscated classes)

Ideally I would move all this constants to actual files, or mock as you are saying

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
violine1101: Java crashes (`hs_err_pid`) are `.log`, Minecraft crashes are all without exception `.txt`

chandler05: It would be nice to copy these deobfuticated crashed to the description for #281, but probably out of scope for this PR.

SPGoding (approved):

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
Marcono1234: Where is the `File` written though? Maybe I am overlooking it, but I did not find where this is happening.

urielsalis: next PR, it was missing 👁️

PR #627, opened 2021-05-11
Update description with crash

### LINKED ISSUE #281: Copy crash report to the description when it is attached

urielsalis:
## Purpose
Fixes #281 

## Approach
image

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
```diff
@@ -0,0 +1,77 @@
...
+                issue.attachments.any { attachment ->
+                    attachment.name == getDeobfName(it.first) || attachment.name.endsWith(
+                        "deobfuscated.txt"
+                    )
+                }
+            }
+        minecraftCrashesWithDeobf.forEach {
+            val file = File(getDeobfName(it.first))
+            file.writeText(it.second!!)
+            issue.addAttachment(file)
```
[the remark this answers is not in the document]
urielsalis: problem is that the function is run after all modules execute, so we cant delete it

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
```diff
@@ -0,0 +1,77 @@
...
+            .filterNot {
+                issue.attachments.any { attachment ->
+                    attachment.name == getDeobfName(it.first) || attachment.name.endsWith(
+                        "deobfuscated.txt"
+                    )
+                }
+            }
+        minecraftCrashesWithDeobf.forEach {
+            val file = File(getDeobfName(it.first))
+            file.writeText(it.second!!)
```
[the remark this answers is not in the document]
violine1101: Perhaps hashing the file name or contents and using that as temporary file name might be the best idea, if we can give a custom name to JIRA still then. Also would be nice to have these temp things in a separate folder.

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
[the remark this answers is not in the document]
Marcono1234: Ah right, nevermind then.

urielsalis:
> Looks good!
> But I think the intention with #281 was partially also to have the crash report easily visible in the description. Having Arisa add a comment might be useful as well, but it might be a different use case. @misode what do you think?
> 
> It might also become spammy when users upload a bunch of crash reports, e.g. "I am not sure which crash report you want so I uploaded all..."
> But we can see if that actually becomes a problem.

Main idea from what I understood is being able to search, as jira doesnt search attachments. Comments are less spammy than description too 👀

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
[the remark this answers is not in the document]
Marcono1234: > Also would be nice to have these temp things in a separate folder.

That is probably pretty important. Ideally we just use a temporary file with random name or directly provide an `InputStream` to Jira Client, but sadly Jira Client does not seem to support providing an attachment name separate from the file name.
So for now either having a dedicated directory or creating a new temporary directory for each file might be best?

Hashing the file name would have the disadvantage that the attachment name on Jira would not be as useful anymore. Removing `/` and `\` is probably fine for now; if we really wanted to be safe, we could also try using approaches similar to the [Zip Slip mitigation](https://snyk.io/research/zip-slip-vulnerability#java) ones.

misode:
In my opinion having the crash report at the bottom of the description would be better. Comments get pushed away most of the time. But now that I say that, editing descriptions also sounds risky, for example the reporter might be in the middle of editing while the module is triggered.

My second question would be what's the point of having a deobfuscated stack trace?

Marcono1234:
> A question I have: what's the point of having a deobfuscated stack trace?

See #606, mainly:
- to make it easier to reason about where an exception occurred
- to make it easier to compare crash reports accross versions (when obfuscation mappings and line numbers change)

misode:
Oh I'm tired. I meant what's the reason for the obfuscated stack? The deobfuscated stack trace should be easier to compare and give you all the information.

urielsalis:
Deobfuscated trace might be wrong

>

src/test/kotlin/io/github/mojira/arisa/modules/CrashModuleTest.kt
```diff
@@ -173,78 +173,15 @@ const val JAVA_CRASH = """#
...
 private val A_SECOND_AGO = NOW.minusSeconds(1)
 
 const val Unconfirmed = "Unconfirmed"
 val NoPriority = null
 
-val crashReader = CrashReader()
+private val crashReader = CrashReader()
 
-class CrashModuleTest : StringSpec({
+class CrashInfoModuleTest : StringSpec({
```
violine1101, line 184:
```suggestion
class CrashModuleTest : StringSpec({
```

src/test/kotlin/io/github/mojira/arisa/modules/CrashInfoModuleTest.kt
```diff
@@ -0,0 +1,197 @@
...
+	Current Language: English (US)
+	CPU: 4x Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz
+"""
+
+private val NOW = Instant.now()
+private val A_SECOND_AGO = NOW.minusSeconds(1)
+
+private val crashReader = CrashReader()
+
+class CrashModuleTest : StringSpec({
```
violine1101, line 83:
```suggestion
class CrashInfoModuleTest : StringSpec({
```

src/test/kotlin/io/github/mojira/arisa/modules/CrashInfoModuleTest.kt
```diff
@@ -0,0 +1,197 @@
...
+        )
+
+        val result = module(issue, A_SECOND_AGO)
+
+        result.shouldBeRight(ModuleResponse)
+        resolvedAsDupe.shouldBeFalse()
+        addedAttachement.shouldBeTrue()
+        resolvedAsInvalid.shouldBeFalse()
+    }
+})
```
violine1101, line 185: No tests for the comment that this PR adds?

violine1101 (commented):
Looks good tech-wise, but I have similar concerns as Misode and Marcono.

I don't think this should be left as a comment. As said in #281 already, I really only think this crash should be integrated into the description when the bug report is initially created.

This is most important if the crash report has been added to the description and the newlines have been messed up. In all other cases it's trivial for a mod or helper to copy stuff from attachments to the description, although it would still be a nice QoL change to not have to open an attachment when checking out a new crash bug report.

Adding a comment every time a crash report is added just causes spam, especially if someone is intentionally uploading unrelated crash reports to a ticket.

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
```diff
@@ -0,0 +1,90 @@
...
+                    attachment.name == getDeobfName(it.first) || attachment.name.endsWith(
+                        "deobfuscated.txt"
+                    )
+                }
+            }
+        minecraftCrashesWithDeobf.forEach {
+            if (isNotZipSlip(getDeobfName(it.first))) {
+                val file = File(crashesDir, getDeobfName(it.first))
+                file.writeText(it.second!!)
+                issue.addAttachment(file)
```
Marcono1234, line 61: Might be good to add a comment to prevent further confusion regarding if and where the file is deleted.
```suggestion
                // Also deletes the file once it has been attached
                issue.addAttachment(file)
```

src/test/kotlin/io/github/mojira/arisa/modules/CrashInfoModuleTest.kt
chandler05: Added tests.

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -344,6 +344,26 @@ fun addRestrictedComment(
     }
 }
 
+fun addRawComment(
```
Marcono1234, line 347: This is not needed anymore, is it?

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
```diff
@@ -0,0 +1,92 @@
...
+) : Module {
+    override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
+        Either.fx {
+            val crashes = AttachmentUtils(crashReportExtensions, crashReader).extractCrashesFromAttachments(issue)
+            val newCrashes = assertContainsNewCrash(crashes, lastRun).bind()
+            uploadDeobfuscatedCrashes(issue, newCrashes)
+
+            newCrashes
+                .filter { it.second is Crash.Minecraft }
+                .filterNot { it.first.name.endsWith("deobfuscated.txt") }
```
Marcono1234, line 28: We should probably restrict this to crash reports attached by the reporter (or by a helper / moderator?).
Otherwise users can vandalize existing issues by attaching unrelated crash reports, or files which Arisa considers to be crash reports.

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
```diff
@@ -0,0 +1,92 @@
...
+                .forEach {
+                    if (!description!!.contains("""\{code.*${it.first.name}]}(\S|\s)*\{code}""".toRegex())) {
```
Marcono1234, line 30: Is `(\S|\s)` to match line breaks? Maybe the complete `(\S|\s)*\{code}` part can be removed, it is rather unlikely that the description contains `{code:title=...}` which is not actually a code block.

Additionally we should probably only add the crash report to the description if there is none, to prevent adding multiple crash report snippets. Therefore we should probably only consider `newCrashes` if its size is 1, otherwise Arisa might pick the 'wrong' crash report in case a user uploads unrelated ones.

The regex could probably changed to this:
```suggestion
                    // Check if description already contains code block which might contain crash report
                    // Note: Jira seems to allow spaces between most of the characters
                    if (!description!!.contains("""\{code:\s*title\s*=.*\[\^.+\.txt\s*]\s*}""".toRegex())) {
```

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
chandler05: It'd be nice to leave it for future purposes.

src/main/kotlin/io/github/mojira/arisa/modules/CrashInfoModule.kt
Marcono1234: Maybe this regex is too unspecific though and causes some false negatives, e.g. when `{code}` blocks link to a non-crash attachment (e.g. code snippet file when a code analysis was done). But I assume that this will happen rarely, so it might be fine for now and we can adjust it later if necessary.

PR #632, opened 2021-05-12
Fix removing both Relates links

### LINKED ISSUE #487: RemoveIdenticalLink module removes both links
## The Bug
RemoveIdenticalLink module removes both links.

## Example Ticket
https://bugs.mojang.com/browse/MC-199657?focusedCommentId=817403&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-817403

## Expected behavior
Only remove one of the two relates links. (Perhaps always the outwards one?)

SPGoding:
## Purpose
Fix #487.

## Approach
By always removing the outwards link belonging to the smaller ticket key.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
@@ -27,5 +27,5 @@ class RemoveIdenticalLinkModule : Module {
         }
     }
 
-    private fun isRelatesLink(link: Link): Boolean = link.type.toLowerCase() === "relates"
+    private fun isRelatesLink(link: Link): Boolean = link.type.toLowerCase() == "relates"
 }
--- a/src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
@@ -3,40 +3,29 @@ package io.github.mojira.arisa.modules
 import arrow.core.Either
 import arrow.core.extensions.fx
 import io.github.mojira.arisa.domain.Issue
+import io.github.mojira.arisa.domain.Link
 import java.time.Instant
 
 class RemoveIdenticalLinkModule : Module {
-    data class GroupingLink(
-        val type: String,
-        val outwards: Boolean,
-        val key: String
-    )
-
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
```diff
@@ -3,40 +3,29 @@ package io.github.mojira.arisa.modules
...
-                .values
-                .flatMap { list ->
-                    list.subList(1, list.size).map { it.remove }
+                .filter(::isRelatesLink)
+                .groupBy { it.issue.key }
+                .filterValues { it.size == 2 }
+                .map {
+                    // Always remove the outwards link belonging to the smaller ticket key,
+                    // so that when this module is triggered on both tickets, only one link is removed.
+                    it.value.find { l -> if (key < it.key) l.outwards else !l.outwards }?.remove
```
chandler05, line 21: I don't think this compares the ticket keys properly, as the key includes the letters and you're trying to compare them.

This doesn't always work properly, for example, "MC-14" is considered less than "MC-2".

I'd recomment comparing after splitting at the dash and converting the leftover string to an int.

SPGoding: It really doesn't matter, we just need some consistent way to compare the two keys :eyes:

urielsalis (dismissed):

src/main/kotlin/io/github/mojira/arisa/modules/RemoveIdenticalLinkModule.kt
```diff
@@ -3,40 +3,29 @@ package io.github.mojira.arisa.modules
...
                 }
 
             assertNotEmpty(removeLinkFunctions).bind()
 
-            removeLinkFunctions.forEach { it.invoke() }
+            removeLinkFunctions.forEach { it?.invoke() }
         }
     }
+
+    private fun isRelatesLink(link: Link): Boolean = link.type.toLowerCase() === "relates"
```
chandler05, line 30:
```suggestion
    private fun isRelatesLink(link: Link): Boolean = link.type.toLowerCase() == "relates"
```

SPGoding: Tysm!

chandler05 (approved):

PR #634, opened 2021-05-13
Enable platform modules on MCD

### LINKED ISSUE #633: Support MCD in the platform modules
## The Problem
MCD uses a different custom field for their platform, so the original keepPlatform and multiplePlatforms modules won't work in that project.
## Possible Solution
Implement identical modules for dungeons, or somehow read both fields in the keepPlatform and multiplePlatforms modules.

SPGoding:
## Purpose
Fix #633.

## Approach
Read the corresponding field according to the issue's project.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


chandler05:
Dungeons has other platforms that need to be added to the config at line 255

src/test/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModuleTest.kt
```diff
@@ -238,7 +238,7 @@ class MultiplePlatformsModuleTest : StringSpec({
         val issue = mockIssue(
```
chandler05, line 238: New dungeons config values need to be added to the test module as well.

SPGoding: I was just doing that xD

src/main/kotlin/io/github/mojira/arisa/modules/KeepPlatformModule.kt
```diff
@@ -14,20 +15,23 @@ class KeepPlatformModule(
...
         }
     }
 
-    private fun isPlatformChange(item: ChangeLogItem) =
-        item.field == "Platform"
+    private fun Issue.getPlatformValue() =
+        (if (project.key == "MCD") dungeonsPlatform else platform).getOrDefault("None")
+
+    private fun isPlatformChange(project: String, item: ChangeLogItem) =
+        item.field == (if (project == "MCD") "Platform" else "Dungeons Platform")
```
chandler05, line 34:
```suggestion
        item.field == (if (project == "MCD") "Dungeons Platform" else "Platform")
```
Looks like this was backwards

SPGoding: Oops, tysm!!

config/config.yml
```diff
@@ -221,6 +222,15 @@ arisa:
         - MCPE
```
chandler05, line 222: Looks like both modules still only run in MCPE, MCD needs to be added to the projects config here and at KeepPlatform's config.

PR #635, opened 2021-05-13
Fix reading linked text in Multiple Platforms

chandler05:


SPGoding:
image

The original message is correct afaik

chandler05:
Ah yep, you are correct. I was going off of what Jira displayed in the changelog, which was wrong. Thank you!

SPGoding (approved):

PR #636, opened 2021-05-13
Fix error when updating Dungeons Platform in MultiplePlatform

SPGoding:
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/test/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModuleTest.kt
+++ b/src/test/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModuleTest.kt
@@ -2,7 +2,12 @@ package io.github.mojira.arisa.modules
 
 import arrow.core.left
 import arrow.core.right
-import io.github.mojira.arisa.utils.*
+import io.github.mojira.arisa.utils.RIGHT_NOW
+import io.github.mojira.arisa.utils.mockIssue
+import io.github.mojira.arisa.utils.mockLink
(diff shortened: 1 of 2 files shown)


chandler05 (approved):

src/test/kotlin/io/github/mojira/arisa/modules/MultiplePlatformsModuleTest.kt
```diff
@@ -2,7 +2,12 @@ package io.github.mojira.arisa.modules
 
 import arrow.core.left
 import arrow.core.right
-import io.github.mojira.arisa.utils.*
+import io.github.mojira.arisa.utils.RIGHT_NOW
+import io.github.mojira.arisa.utils.mockIssue
+import io.github.mojira.arisa.utils.mockLink
+import io.github.mojira.arisa.utils.mockLinkedIssue
+import io.github.mojira.arisa.utils.mockComment
+import io.github.mojira.arisa.utils.mockChangeLogItem
```
Marcono1234: In IntelliJ you can change the Code Style settings to prevent this, see [this Stack Overflow question](https://stackoverflow.com/questions/48755419/disable-intellij-kotlin-starred-imports) (in case that was the cause)

SPGoding: Thank you, though I believe I have changed that option multiple times now and it just keeps resetting itself.

PR #467, opened 2020-10-18
Create Thumbnail Module (Needs takeover)

SunCatMC:
## Purpose
Resolve https://github.com/mojira/arisa-kt/issues/165

## Approach
Have the module run only for newly created comments and tickets, add |thumbnail to images uploaded to jira and embedded in the ticket/comment, and add set width and height to images linked from outside the jira (as [per jira](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=images) thumbnails don't work for those; see also [my own test](https://bugs.mojang.com/browse/MCTEST-110?focusedCommentId=826663&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-826663))

#### Checklist
- [ ] Included tests
- [ ] Updated documentation in [README](https://github.com/mojira/arisa-kt/blob/master/README.md) and [wiki](https://github.com/mojira/arisa-kt/wiki)
- [ ] Tested in MCTEST-xxx

PR #640, opened 2021-05-13
Add ThumbnailModule for attachments

### LINKED ISSUE #165: Add |thumbnail to large images
## The Problem
It happens now and then that users attach a large image and then in their comment or report display that image (`!image.jpg!`). This makes the image take up too much space and sometimes also cuts it off (depending on the size of the screen you are viewing it), requiring horizontal scrolling.
It would therefore be good to add `|thumbnail` to the image: `!image.jpg|thumbnail!`

What do you think?

## Considerations
- The module would have to download the image and check its size. When determining which image formats Arisa should support, we would need to check:
    - which image formats are supported by Jira
    - for which image formats this problem occurs most, probably PNG. Supporting other images formats is probably not worth it.
- The module should only run for attached images, and not links to external sites. Otherwise that would reveal Arisa's IP address. This is probably alright because it appears this problem of too large images occurrs mostly for attached images anyways. We need to check:
    - how Jira determines whether an image link points to an attachment or to an external site (e.g. `!example.com!`)
- The module should not trigger if the user specified attributes for the image (i.e. `|` is present).
- The module should not download images too often to not decrease performance too much, e.g. in case a user frequently edits a report or comment. Maybe the module should only consider the initial comment / report description and any edits which occur within one day (or a shorter time span) after creation of the comment / report.

Marcono1234:
## Purpose
Resolves #165
Supersedes  #467

## Approach
When the module detects embedded images (e.g. `!test.png!`) in the description of newly created issues or in newly created comments it reads the image using [Apache Commons Imaging](https://commons.apache.org/proper/commons-imaging/) (as described in https://github.com/mojira/arisa-kt/issues/165#issuecomment-632961412) to determine the size of the image. If the image is wider or higher than a certain maximum `|thumbnail` is added to the image reference.

The module currently only checks for attached images, but not for URL image references, as originally intended by #467. Handling URL references is slightly more complex because `|thumbnail` does not work. Additionally a malicious user might be able to trigger a DoS attack (similar to [Slowloris](https://en.wikipedia.org/wiki/Slowloris_(computer_security))), or might obtain information about the server running Arisa, such as the IP address (might not be a problem). However, it appears the majority of embedded images references attachments, so handling URL references is probably not worth the effort.

Note that I am not very familiar with Kotlin, so the code might not follow best practices. Therefore I have marked this pull request as Draft for now, would be great if @urielsalis or someone else with more experience could have a look.

## Open questions / tasks
- Should the maximum image size be configurable in the config? (probably)
Will add a commit for this then.
- Editing the description of an issue or a comment directly after it was created might cause a race condition when the user directly makes an edit afterwards and Arisa overwrites it (likelyhood for this might be increased due to delay for downloading and reading image). Could be avoided / reduced by only acting _X_ minutes after creation, but that would not work with Arisa's current JQL for selecting relevant issues when no other update for that issue occurred.
- Not tested in MCTEST yet (don't have the test setup for this yet either)

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


build.gradle.kts
```diff
@@ -60,6 +61,9 @@ dependencies {
     implementation("io.arrow-kt", "arrow-fx", arrowVersion)
     implementation("com.beust", "klaxon", "5.4")
     implementation("com.mojang", "brigadier", "1.0.17")
+    implementation("io.ktor", "ktor-client-core", ktorVersion)
+    implementation("io.ktor", "ktor-client-cio", ktorVersion)
+    implementation("org.apache.commons", "commons-imaging", "1.0-alpha2")
```
Marcono1234, line 66: Uses alpha version because there exists no release yet.

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
```diff
@@ -0,0 +1,189 @@
...
+                val newBody = replaceEmbeddedImages(issue, oldBody)
+                if (oldBody != newBody) {
+                    it.update(newBody)
+                    performedUpdate = true
+                }
+            }
+
+            // TODO: Not sure if this is correct; should assert be called before
+            // modifying issue?
+            assertTrue(performedUpdate).bind()
```
Marcono1234, line 120: Could someone more familiar with Kotlin / Arrow answer that?
(`TODO` comment also causes the build failure)

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
```diff
@@ -0,0 +1,189 @@
...
+            newComments.forEach {
+                val oldBody = it.body!!
+                val newBody = replaceEmbeddedImages(issue, oldBody)
+                if (oldBody != newBody) {
+                    it.update(newBody)
+                    performedUpdate = true
+                }
+            }
+
+            // TODO: Not sure if this is correct; should assert be called before
```
chandler05, line 118:
```suggestion
            // TODO: Not sure if this is correct, should assert be called before
```
Detekt getting angered over the semicolon here.

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
chandler05: Probably because of the semicolon, see below.

build.gradle.kts
```diff
@@ -60,6 +61,9 @@ dependencies {
     implementation("io.arrow-kt", "arrow-fx", arrowVersion)
     implementation("com.beust", "klaxon", "5.4")
     implementation("com.mojang", "brigadier", "1.0.17")
+    implementation("io.ktor", "ktor-client-core", ktorVersion)
```
urielsalis, line 64: Instead of adding the full ktor library just for a call, I would use HttpUrlConnection like you do in java, ktor its pretty big and they break their API all the time

urielsalis: and we can modify the download function instead to get a input stream

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
```diff
@@ -0,0 +1,189 @@
...
+ * Note that the exact value does not matter since Jira can be scrolled vertically; though
+ * the value should not be too large nonetheless.
+ */
+const val MAX_HEIGHT = 600
+
+class ThumbnailModule(private val maxImagesCount: Int) : Module {
+    /**
+     * [InputStream] which reads at most a certain number of bytes.
+     */
+    class LimitingInputStream(private val stream: InputStream, private val maxBytes: Long) : InputStream() {
```
urielsalis, line 40: I would move this to a separate file

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
urielsalis: IIRC we only save the updates and execute them after all modules have ran, so description wouldnt be updated here

urielsalis: apart from that, I think its fine.

Marcono1234:
Rebased onto `master`, switched to JDK's Http Client and extracted ThumbnailModule settings to the config.
Could you please review the changes again?

build.gradle.kts
```diff
@@ -33,6 +33,7 @@ configurations {
 
 val logBackVersion = "1.2.3"
 val arrowVersion = "0.10.4"
+val ktorVersion = "1.5.4"
```
urielsalis, line 36: No longer needed

src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
Marcono1234: Thanks. I only added that as temporary comment to clear whether usage of assert is correct there. Have removed the comment now.

build.gradle.kts
Marcono1234: Thanks! Have fixed it now.

urielsalis (approved):

PR #644, opened 2021-05-15
Remove required whitespace in commands, change command prefix, remove non-helper ARISA tags and commands

### LINKED ISSUE #587: ARISA_FIX_CAPITALIZATION does not validate argument
## The Bug
The command `ARISA_FIX_CAPITALIZATION` has an argument which is not validated at all. The [wiki](https://github.com/mojira/arisa-kt/wiki/Commands#arisa_fix_capitalization) says the command is `ARISA_FIX_CAPITALIZATION description`, but you can actually run it with any argument, e.g.:
```
ARISA_FIX_CAPITALIZATION does_not_matter_at_all
```

## Expected behavior
It would be good to validate that the argument is `description` and otherwise thrown an exception.
Otherwise one might use `ARISA_FIX_CAPITALIZATION summary` and then wonder why the command fails, or why it updated the description.

### LINKED ISSUE #614: `RemoveNonStaffMeqsModule` does not cover `ARISA_` commands and tags
## The Bug
The `RemoveNonStaffMeqsModule` currently only covers `MEQS_` tags, however it does not cover `ARISA_` commands, or the `ARISA_REOPEN_OP` tag.

## Expected behavior
Either that module or a different module should also remove `ARISA_` commands and tags.

Maybe we could also restructure the config to have one object listing all tags (`ModuleRegistry` would then get the tag values from there), e.g.:
```yml
tags:
  - keepPrivate: MEQS_KEEP_PRIVATE
  - ...
```
This would reduce duplication of tags we currently have in the config and would allow the `RemoveNonStaffMeqsModule` to get the relevant prefixes (everything up to including the first `_`).
Similarly the `RemoveNonStaffMeqsModule` should probably respect the `command.commandPrefix` value.

chandler05:
## Purpose
Fix #614 and Fix #587 
## Approach
Change command prefix to `$ARISA`, rename `RemoveNonStaffMeqsModule` to `RemoveNonStaffTagsModule` and add Arisa tags and commands to it
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis (approved):
Going to work on the delete Arisa command?

src/main/kotlin/io/github/mojira/arisa/modules/RemoveNonStaffTagsModule.kt
```diff
@@ -0,0 +1,53 @@
...
+    override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
+        Either.fx {
+            val updateMeqsComments = comments
+                .filter(::hasMeqsTag)
+                .filter(::isNotStaffRestricted)
+                .map { it.restrict.partially1(removeTags(it.body!!, """(MEQS)(_[A-Z_]+)""".toRegex())) }
+            val updatePrefixedComments = comments
+                .filter(::hasPrefixedTag)
+                .filter(::isNotVolunteerRestricted)
+                .map { it.restrict.partially1(removeTags(it.body!!, """($removePrefix)(_[A-Z_]+)""".toRegex())) }
```
Marcono1234, line 20: Maybe to prevent code duplication for `has...()` and `removeTags()`, just call `removeTags` and then filter all out all comment bodies where removing tags had no effect. There there would be no need for a separate `has...()` function.

Marcono1234 (commented):
Maybe it would be more flexible to instead have a `removePrefixes` list which then also contains `MEQS`, instead of hardcoding it here.
All comments could then be filtered using `isNotVolunteerRestricted`, though this would allow helper restricted messages containing MEQS tags to remain, but I am not sure how problematic that would be.

src/main/kotlin/io/github/mojira/arisa/modules/RemoveNonStaffTagsModule.kt
chandler05: I've refactored the module to include the list of prefixes, but I don't think it's necessary to remove the `hasPrefixedTag()` function, however in the refactor there is only a single function for this.

Marcono1234:
This pull request is not complete yet, is it? #587 has not been fixed, right?

chandler05:
I removed the need for an argument in that command whatsoever

Marcono1234:
You mean the requirement for a space in the command, right?
Though #587 was mainly about the argument not being validated, and it looks like `CommandDispatcher` still has an unused argument:
https://github.com/mojira/arisa-kt/blob/5d24d12af7ea7392a538bdc6d4d9716e56f2787b/src/main/kotlin/io/github/mojira/arisa/modules/commands/CommandDispatcher.kt#L106-L111

chandler05:
Oops, you are right I forgot to remove it there. Fixed now 👍

urielsalis:
The idea is that we can remove th argument entirely instead of validating it

Marcono1234:
> Going to work on the delete Arisa command?

@urielsalis, what did you mean by that, or was it only a question about the scope of this pull request?

urielsalis:
Mostly to see if it was going to be done in this PR too, but it can be
separate

src/main/kotlin/io/github/mojira/arisa/modules/RemoveNonStaffTagsModule.kt
```diff
@@ -0,0 +1,50 @@
...
+    private fun isNotVolunteerRestricted(comment: Comment) =
+        comment.visibilityType != "group" ||
+                !listOf("staff", "global-moderators", "helper").contains(comment.visibilityValue)
+
+    private fun removeTags(comment: String): String {
+        var newComment = comment
+        removablePrefixes.forEach { prefix ->
+            val regex = """($prefix)(?!_REMOVED)(_[A-Z_]+)""".toRegex()
+            newComment = regex.replace(newComment) {
+                "${it.groupValues[1]}_ARISA_REMOVED${it.groupValues[2]} Removal Reason: $removalReason"
```
Marcono1234, line 45: Not that important, but maybe it would be good to make the result easier to read (unless we want to include the tag `ARISA_REMOVED` to make it searchable). The current output looks somewhat weird, e.g. the output of one of the tests:
> ARISA_ARISA_REMOVED_KEEP_AR Removal Reason: Lorem Ipsum. I like QC.

What about this (would require adjusting the tests):
```suggestion
                "Arisa removed prefix '${it.groupValues[1]}' from '${it.groupValues[2]}'; removal reason: $removalReason"
```

Marcono1234 (approved):
The changes look good to me, I also like the `$` prefix for commands, making it easier to differentiate them from regular tags. However, it would be good to inform the others in the internal communication channel once the changes are merged.

Marcono1234 (approved):

urielsalis (approved):

PR #645, opened 2021-05-15
ReopenAwaitingModule: Ignore comments from helpers

violine1101:
## Purpose
Helpers usually don't want a ticket to be reopened when they post a comment on an AR tickets. Thus they should probably be ignored.

## Approach
Add `helper` to `blacklistedRoles` config

## Future work
N/A


chandler05:
If we add this then helpers can never reopen AR tickets at all.

urielsalis (approved):

bemoty:
as mark has said on slack: we might want to add a command for reopening issues instead (e. g. `ARISA_REOPEN`)

PR #646, opened 2021-05-15
Update Detekt plugin

Marcono1234:
## Purpose
Update Detekt plugin to a newer version.

Ideally we should also update the ktlint plugin, but it looks like the new version has a different formatting style and finds a lot of violations.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     kotlin("jvm") version "1.4.30"
     id("org.jlleitschuh.gradle.ktlint") version "9.2.1"
     application
-    id("io.gitlab.arturbosch.detekt") version "1.9.1"
+    id("io.gitlab.arturbosch.detekt") version "1.17.0"
     id("info.solidsoft.pitest") version "1.5.1"
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -91,7 +84,9 @@ application {
 }
 
 detekt {
-    failFast = true // fail build on any finding
+    // For now only consider main source files, but ignore test sources
+    input = files("src/main/kotlin")
```
Marcono1234, line 88: Detekt now by default also checks test sources (see https://github.com/detekt/detekt/pull/3649), that causes multiple findings for Arisa due to too long lines and too large classes, so for now it might be best to not check test sources.

Maybe in the future we could [configure](https://detekt.github.io/detekt/configurations.html) Detekt to disable checks such as `style` in test sources instead, but still run the other checks (to detect broken tests).

build.gradle.kts
```diff
@@ -91,7 +84,9 @@ application {
...
+    allRules = true // enable all rules, including unstable ones
```
Marcono1234, line 89: `failFast` has been deprecated, see https://github.com/detekt/detekt/pull/3431. When using it, Detekt recommends to use `allRules = true`, since that was apparently the previous behavior of it as well.

urielsalis (approved):

PR #649, opened 2021-05-16
Revert to using HttpURLConnection for language API call

Marcono1234:
## Purpose
Reverts the LanguageDetectionApi.kt changes done by 02bca459eaf707efd6819ca727c57c0e9ebac2cf. For some reason the requests did not work properly and the server complained about missing token value.

E.g. when using an invalid token the error response was:
```
{"error":true,"status":403,"code":"error.notAllowed","message":"token param is missing","data":{"parameter":"token"}}
```

Now with the reverted changes the error response is:
```
{"error":true,"status":401,"code":"error.invalidParameter","message":"no such token 'abcdef'","data":{"parameter":"token"}}
```

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/LanguageDetectionApi.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/LanguageDetectionApi.kt
@@ -2,12 +2,14 @@ package io.github.mojira.arisa.infrastructure
 
 import arrow.core.Either
 import com.beust.klaxon.Klaxon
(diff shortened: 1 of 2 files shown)


urielsalis (approved):

Marcono1234:
It looks like the underlying reason is that by default `HttpURLConnection` sets the header `Content-Type: application/x-www-form-urlencoded`.
The JDK Http Client does not do that, which is why the requests were failing. We could add that header and switch back to JDK Http Client in the future, but for now `HttpURLConnection` should probably work fine.

PR #650, opened 2021-05-16
Improve language detection API logging

Marcono1234:
## Purpose
Improve logging of language API requests.

## Approach
Depends on #649

When logging when calling the language API:
- Log quota information, see https://dandelion.eu/docs/api/#api-response-headers
- Log error response body
:warning: This might reveal sensitive information; for example when the token is invalid it looks like the reponse contains the token. Maybe that is not desired?

Also changes the `LanguageDetectionApi.kt` function to throw exceptions instead of using `Either`. Previously the language module would siltently ignore `Either.left`, without logging the exception. Additionally switches from `Error` to `IllegalArgumentException` / `IOException`; `Error` is usually only used for really critical exceptions, see [documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Error.html).

Please note that have only tested this with invalid tokens yet because I don't have a valid language API token.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
@@ -20,7 +20,7 @@ class ThumbnailModule(
     private val maxImagesCount: Int
 ) : Module {
 
-    private val log: Logger = LoggerFactory.getLogger(ThumbnailModule::class.java)
+    private val log: Logger = LoggerFactory.getLogger("ThumbnailModule")
 
     override fun invoke(issue: Issue, lastRun: Instant): Either<ModuleError, ModuleResponse> = with(issue) {
         Either.fx {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/ThumbnailModule.kt
```diff
@@ -20,7 +20,7 @@ class ThumbnailModule(
     private val maxImagesCount: Int
 ) : Module {
 
-    private val log: Logger = LoggerFactory.getLogger(ThumbnailModule::class.java)
+    private val log: Logger = LoggerFactory.getLogger("ThumbnailModule")
```
Marcono1234, line 23: Changed this to be consistent with other classes which also only use the name and not the fully qualified class name.

urielsalis (approved):

PR #652, opened 2021-05-16
Improve helper messages fetching

Marcono1234:
## Purpose
Improve fetching, parsing and storing of helper messages. Most importantly:
- Fix the following logic being wrong, writing the helper messages before they have been deserialized:
https://github.com/mojira/arisa-kt/blob/02bca459eaf707efd6819ca727c57c0e9ebac2cf/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt#L135-L136
- Fix no exceptions being logged / start not failing without helper messages; previously `updateHelperMessages(...)` returned an `Either<Error, ...>` but the caller never inspected the result. Now instead throw exceptions instead of using `Either`.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
@@ -5,17 +5,21 @@ import arrow.core.left
 import arrow.core.right
 import arrow.core.rightIfNotNull
 import com.beust.klaxon.Klaxon
+import com.beust.klaxon.KlaxonException
 import io.github.mojira.arisa.modules.openHttpGetInputStream
+import org.slf4j.LoggerFactory
 import java.io.File
 import java.io.IOException
 import java.io.InputStream
 import java.net.URI
 
-typealias ProjectFilter = Any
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
```diff
@@ -116,36 +121,48 @@ object HelperMessageService {
...
-                it.left()
-            }
-        },
-        { inputStream ->
-            inputStream.use {
-                file.writeText(data.toJSON())
-                data.fromStream(it).rightIfNotNull {
-                    Error("Couldn't deserialize downloaded helper messages")
+    fun setHelperMessages(json: String) = data.fromJSON(json)
+        ?: throw IOException("Couldn't deserialize helper messages from setHelperMessages()")
```
Marcono1234, line 125: Maybe that is not even needed, it appears Klaxon throws exceptions for invalid JSON, so maybe result is only `null` when input is JSON `null`?

urielsalis (approved):

PR #658, opened 2021-05-17
Set user-agent for language api request

chandler05:
## Purpose
Potentially fix 403 errors
## Approach
https://stackoverflow.com/questions/3869372/java-io-ioexception-server-returned-http-response-code-403-for-url and https://stackoverflow.com/questions/16826345/why-do-i-get-a-403-error-when-i-try-open-a-url

I followed these two answers, which I hopefully did right.


src/main/kotlin/io/github/mojira/arisa/infrastructure/LanguageDetectionApi.kt
```diff
@@ -60,6 +60,11 @@ fun getLanguage(token: String?, text: String): Map<String, Double> {
 
     val request = "token=" + URLEncoder.encode(token, UTF_8) + "&text=" + URLEncoder.encode(text, UTF_8)
     with(URL(BASE_URL).openConnection() as HttpURLConnection) {
+        this.setRequestProperty(
```
urielsalis, line 63: Wait, also add this.setRequestProperty("Content-Type", "application/json")

urielsalis (approved):

PR #667, opened 2021-05-21
Use the normal duplicate comment instead of private if the reporter is the same

### LINKED ISSUE #666: If reporter of a private duplicate is the same as the parent ticket, don't use the private duplicate message
## The Problem
Arisa uses the private duplicate message in this case, however it is incorrect in saying the reporter cannot access the parent ticket in this situation
## Possible Solution
Use the normal duplicate message.

chandler05:
## Purpose
Fix #666
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/DuplicateMessageModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/DuplicateMessageModule.kt
@@ -54,7 +54,7 @@ class DuplicateMessageModule(
                 val fullParents = fullParentEitherList
                     .map { (it as Either.Right).b }
 
-                messageKey = getPrivateMessageOrNull(fullParents)
+                messageKey = getPrivateMessageOrNull(issue, fullParents)
                     ?: getResolutionMessageOrNull(fullParents) ?: getForwardResolvedOrNot(issue, fullParents)
             }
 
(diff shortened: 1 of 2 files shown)


urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/DuplicateMessageModule.kt
```diff
@@ -94,9 +94,12 @@ class DuplicateMessageModule(
             getField(this[0])
         }
 
-    private fun getPrivateMessageOrNull(issues: List<Issue>): String? {
+    private fun getPrivateMessageOrNull(child: Issue, issues: List<Issue>): String? {
         val parentSecurity = issues.getCommonFieldOrNull { it.securityLevel }
-        return privateMessage.takeIf { parentSecurity != null }
+        val reporters = issues.getCommonFieldOrNull { it.reporter }
+        return privateMessage.takeIf {
+            parentSecurity != null && (reporters == null || child.reporter != reporters)
```
SPGoding, line 101: Simply `reporters == null` should be enough iirc. The field will be `null` when the reporters on those ticket are not the same.

SPGoding (approved):

PR #675, opened 2021-05-31
Use threshold for logging language API quota warning

Marcono1234:
## Purpose
As mentioned by @urielsalis internally the current quota logging for the language API is too spammy because it is logged for every request.

## Approach
Define a `apiQuotaWarningThreshold` config entry. When the quota units become equal to or lower than this value a warning will be logged once.
Therefore no further messages are logged while the quota remains that low, and only once it was reset and again became lower than the threshold a warning will be logged again.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/config/ArisaConfig.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/config/ArisaConfig.kt
@@ -166,6 +166,11 @@ object Arisa : ConfigSpec() {
         }
 
         object Language : ModuleConfigSpec() {
+            val apiQuotaWarningThreshold by optional<Double?>(
+                description = "When the number of remaining API quota units becomes equal to or lower than this" +
+                        " value a warning will be logged.",
+                default = null
+            )
             val allowedLanguages by required<List<String>>(
                 description = "Codes of languages that can be used."
             )
--- a/config/config.yml
+++ b/config/config.yml
@@ -377,7 +377,7 @@ arisa:
         - Unresolved
       excludedStatuses:
         - Postponed
-      apiQuotaWarningThreshold: 15.0
+      apiQuotaWarningThreshold: 10.0
       allowedLanguages:
         - en
         - nl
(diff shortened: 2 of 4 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/config/ArisaConfig.kt
```diff
@@ -166,6 +166,11 @@ object Arisa : ConfigSpec() {
         }
 
         object Language : ModuleConfigSpec() {
+            val apiQuotaWarningThreshold by optional<Double?>(
+                description = "When the number of remaining API quota units becomes equal to or lower than this" +
+                        " value a warning will be logged.",
+                default = null
```
urielsalis, line 172: Let's set it to 10 by default? (That should be around 100 tickets

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/infrastructure/config/ArisaConfig.kt
Marcono1234: Sure! Also feel free to squash these commits when merging.

Marcono1234: Ah, sorry misunderstood your comment. Do you mean change the default here?
I thought it might be useful to allow disabling the logging (for whatever reason) by simply not specifying a value in the config.
Do you think the threshold should be non-`null` then?

urielsalis: Yep

urielsalis (approved):

Marcono1234:
@urielsalis should I create a follow-up pull request making the threshold non-`null` as requested in your review comment above?

urielsalis:
no need, as long as its set to 10 in config.yml its fine by me

PR #681, opened 2021-06-05
Fix #680 - Append deobf to the start instead of the end of attachment name

### LINKED ISSUE #680: Name deobfuscated crash reports with deobfuscated stated at the beginning rather than the end
## The Problem

Currently, it is very hard to distinguish obfuscated crash reports from crash reports Arisa deobfuscated:
grafik

You have to hover over the name to see the "deobfuscated" part of the file name: 
grafik

## Possible Solution

Name the crash report `deobf_<original-name>` instead.

urielsalis:
## Purpose
Fix #680
## Approach
Append deobf at the start
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
@@ -58,7 +58,9 @@ class CrashModule(
             .filter { it.second != null }
             .filterNot {
                 issue.attachments.any { attachment ->
-                    attachment.name == getDeobfName(it.first) || attachment.name.startsWith("deobf_")
+                    attachment.name == getDeobfName(it.first) ||
+                        attachment.name.startsWith("deobf_") ||
+                        attachment.name.endsWith("deobfuscated.txt")
                 }
             }
         minecraftCrashesWithDeobf.forEach {
--- a/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
@@ -58,9 +58,7 @@ class CrashModule(
             .filter { it.second != null }
             .filterNot {
                 issue.attachments.any { attachment ->
-                    attachment.name == getDeobfName(it.first) || attachment.name.endsWith(
-                        "deobfuscated.txt"
-                    )
+                    attachment.name == getDeobfName(it.first) || attachment.name.startsWith("deobf_")
                 }
             }
         minecraftCrashesWithDeobf.forEach {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
```diff
@@ -58,9 +58,7 @@ class CrashModule(
             .filter { it.second != null }
             .filterNot {
                 issue.attachments.any { attachment ->
-                    attachment.name == getDeobfName(it.first) || attachment.name.endsWith(
-                        "deobfuscated.txt"
-                    )
+                    attachment.name == getDeobfName(it.first) || attachment.name.startsWith("deobf_")
```
violine1101, line 61: Not really an issue with this PR, but as I understand it this means that Arisa will only ever deobfuscate one crash report on each ticket? Is that desired?

urielsalis: It will only deobfuscate it once, we have tickets with more than 1 crash and it deobfuscates all of them

this check makes sure we dont try to deobfuscate a crash that has already been deobfuscated (To not have a infinite loop)

violine1101: I'm only talking about the second part here, it checks if there's _any_ attachment which is a deobfuscated crash, hence Arisa can only deobfuscate anything on a ticket once.

We already make sure that we only analyze new crashes in line 33 above. IMO the second part of this check is unnecessary or even a bug.

violine1101: Ah, wait, I think I see now what the issue is here. But the implementation is faulty.

We should check whether the attachment starts with `deobf` _outside_ of the lambda inside of inside of it. However perhaps it's more foolproof to check if the uploader is Arisa itself.

violine1101: Will open a new PR for this issue. It doesn't really cause too much trouble here.

violine1101 (approved):

src/main/kotlin/io/github/mojira/arisa/modules/CrashModule.kt
Marcono1234: > hence Arisa can only deobfuscate anything on a ticket once

I would see it like this as well, that is probably not desired. `assertContainsNewCrash` should really suffice, but a safeguard checking whether Arisa uploaded the crash would not hurt (to prevent endless loops in case the creation date is incorrect).
Do you want to open a new issue / pull request for this?

violine1101: Was working on this already, here's the new PR: #684

PR #683, opened 2021-06-05
Fix #679 - Wait 1 minute and retry logging in on jira issues

### LINKED ISSUE #679: Ignore errors and throttle calls to jira when its down
## The Bug
https://canary.discord.com/channels/647810384031645728/728157703398227978/850339444062027777

urielsalis:
## Purpose
On exceptions, wait for 1 minute an relog
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
@@ -1,6 +1,5 @@
 package io.github.mojira.arisa
 
-import arrow.syntax.function.partially1
 import ch.qos.logback.classic.AsyncAppender
 import ch.qos.logback.classic.LoggerContext
 import com.github.napstr.logback.DiscordAppender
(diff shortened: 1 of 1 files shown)


Marcono1234:
Don't we have this (or similar logic) here already:
https://github.com/mojira/arisa-kt/blob/8bdd886cbae93a46308978083771a12c75076cd9/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt#L86-L87

However it looks like this is flawed:
- Due to using `isAfter` it seems to check if the last relog was within the last minute; it might have to be check `isBefore` instead?
- `lastRelog` is not updated after a relog attempt was performed

urielsalis:
That might be what is preventing it from working

urielsalis:
This is high priority, arisa doesnt recover from crashes without it.

@Marcono1234 can you implement those changes in this PR?

PR #689, opened 2021-06-09
Rewrite module execution stack

violine1101:
## Purpose
I found an error in the module execution that might have caused Arisa to check on tickets more often than it should have.

Since the whole thing was kinda messy, I decided to rewrite it altogether.

Please review carefully and give feedback.

## Approach
I split up the registry into three separate registries. Each of these registries has a fixed jql query. For each registry, all tickets get queried at once. This means, Arisa will no longer have to query all tickets for each module; the query cache is redundant and has been removed.

The module executor subsequently has been split up as well, into the "Executor" for handling the high level stuff, and the "ModuleExecutor", for only executing a single module.

## Future work
Documentation is still missing. Also I don't like how I massacred the tests for the ModuleExecutor, however I'm not really sure how to properly test this. I'm not too familiar with mockk either and don't want to fight with that right now.

In the further future, it might be nice to get rid of the registries altogether and include the registration of the modules directly in the module files ...

#### Checklist
- [ ] Included tests
- [x] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/registry/LazyModuleRegistry.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.github.mojira.arisa.infrastructure.config.Arisa
+import io.github.mojira.arisa.modules.UpdateLinkedModule
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+/**
+ * This class is the registry for modules that get executed "lazily", i.e. with a delay of multiple hours.
+ * This is done in order to avoid spam.
+ */
+class LazyModuleRegistry(config: Config) : ModuleRegistry(config) {
```
Marcono1234, line 13: "Delayed" instead of "lazy" might be more fitting.

However this registry is currently specifically created for `UpdateLinkedModule`, and also has a `updated > ${lastRun.toEpochMilli()}` check which makes at least setting Linked to 0 non-delayed.
Maybe such a generic name as `LazyModuleRegistry` is therefore misleading.

src/main/kotlin/io/github/mojira/arisa/registry/ModuleRegistry.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    private fun getModuleResult(moduleName: String, module: Module) = { issue: Issue, lastRun: Instant ->
+        moduleName to tryExecuteModule { module(issue, lastRun) }
+    }
+
+    private fun getJqlWithDebug(lastRun: Instant): String {
+        val debugWhitelist = config[Arisa.Debug.ticketWhitelist]
+
+        return if (debugWhitelist == null) getJql(lastRun)
+        else "key IN (${debugWhitelist.joinToString()}) AND (${getJql(lastRun)})"
```
Marcono1234, line 70: Maybe store `getJql(lastRun)` in a variable and then use its value instead to make it more obvious that this is always considered (instead of having it duplicated as it is currently the case).

src/main/kotlin/io/github/mojira/arisa/registry/LazyModuleRegistry.kt
violine1101: `DelayedModuleRegistry` exists already; it's used for the `DuplicateMessageModule`, which is run slightly delayed (a few minutes only).

Yeah, the name is probably not optimal. These three registries are just kind of a workaround for different requirements by the different modules.

IMO it would probably make sense to either split the part of the `UpdateLinkedModule` that only sets Linked to 0 into a separate module that would only be responsible for that, or to not have that part at all and have Linked set to 0 by default on JIRA side if that's possible to set up.

src/main/kotlin/io/github/mojira/arisa/Executor.kt
```diff
@@ -0,0 +1,91 @@
...
+            log.debug("Returned issues for registry ${ registry::class.simpleName }: ${ issues.map { it.key } }")
+        } else {
+            log.debug("${ issues.size } issues have been returned for registry ${ registry::class.simpleName }")
+        }
+
+        return issues
+    }
+
+    init {
+        val enabledModules = registries.flatMap { smth -> smth.getEnabledModules().map { it.name } }
```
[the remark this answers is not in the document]
violine1101: Whoops, yeah I intended to change that but forgot about it. Thanks!

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -172,11 +163,10 @@ private fun searchIssues(
             log.warn("Failed to connect to jira. Caused by: ${e.cause?.message}")
             throttledLog = System.currentTimeMillis()
```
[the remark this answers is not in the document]
violine1101: Actually I'm not sure if this ever did anything. It seems like this catch only activates if the jql query is incorrect.

src/main/kotlin/io/github/mojira/arisa/registry/LazyModuleRegistry.kt
violine1101: Renamed it to `LinkedModuleRegistry`

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -57,33 +57,38 @@ fun main() {
...
+    val moduleRegistries = getModuleRegistries(config)
+    var executor = Executor(
+        config, moduleRegistries, getSearchIssues(jiraClient, config)
     )
 
     while (true) {
-        // save time before run, so nothing happening during the run is missed
-        val curRunTime = Instant.now()
-        val executionResults = moduleExecutor.execute(lastRunTime, rerunTickets)
+        // Save time before run, so nothing happening during the run is missed
```
urielsalis, line 67: Not for this PR, but a lot of this logic should be moved out of Main to its own service, Main should be as simple as possible and only orchestrating

src/main/kotlin/io/github/mojira/arisa/Executor.kt
```diff
@@ -0,0 +1,97 @@
...
+            moduleExecutor.executeModule(issues, addFailedTicket) { issue -> execute(issue, timeframe.lastRun) }
+        }
+    }
+
+    private fun getIssuesForRegistry(
+        registry: ModuleRegistry,
+        rerunTickets: Collection<String>,
+        timeframe: TicketQueryTimeframe
+    ): List<Issue> {
+        val issues = mutableListOf<Issue>()
```
urielsalis, line 62: same here, this probably should go to a issue service

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
violine1101: Yeah, it was already kinda tempting to put the `searchIssues` down below somewhere else, but I didn't want to do _too_ many intrusive changes here (especially since marcono's PR also exists).

urielsalis:
Looks good to me, @Marcono1234 whatdo you think?

urielsalis (dismissed):

src/main/kotlin/io/github/mojira/arisa/registry/ModuleRegistry.kt
```diff
@@ -63,19 +92,20 @@ abstract class ModuleRegistry(private val config: Config) {
...
 
-    fun getFullJql(lastRun: Instant, failedTickets: Collection<String>): String {
-        val projects = config[Arisa.Issues.projects]
-        val failedTicketsJql = if (failedTickets.isEmpty()) "key in (${ failedTickets.joinToString() })" else ""
-        val registryJql = "(${ getJqlWithDebug(lastRun) })"
+    fun getFullJql(timeframe: TicketQueryTimeframe, failedTickets: Collection<String>): String {
+        val failedTicketsJql = if (failedTickets.isEmpty()) "" else "key in (${ failedTickets.joinToString() }) OR "
+        val registryJql = "(${ getJqlWithDebug(timeframe) })"
 
-        return "project IN (${ projects.joinToString() } AND ($failedTicketsJql OR $registryJql) ORDER BY updated ASC"
```
Marcono1234: Did you remove `project IN ...` on purpose?

violine1101: Ah, yes, I did. Reason for this is that this caused incorrect behaviour; Arisa also checks tickets that are not in the global config, weirdly enough. Most noticeably, this would have completely disabled the ResolveTrashModule.

violine1101: Note that I added it in the previous commit(s), it was not originally there.

Marcono1234: Ah thanks for the clarification. It looks like module specific project lists take precedence:
https://github.com/mojira/arisa-kt/blob/253faaa191b0c3b34a23493e28a09661a9baf54c/src/main/kotlin/io/github/mojira/arisa/ModuleExecutor.kt#L114

Marcono1234 (approved):
The changes look really good!

I think one minor issue is that the search query of a module registry would probably pick up the changes the modules of a previous registry made in the same run. I assume we had the same issue with the previous implementation as well (unless the JQL queries were the same and were served from the cache). Maybe that was also one of the reasons why catching up took so long?
However, I don't think that this can be easily solved and it is likely not a big issue either with these changes now.

PR #690, opened 2021-06-09
Fix and improve Mojira relogging

### LINKED ISSUE #679: Ignore errors and throttle calls to jira when its down

Marcono1234:
## Purpose
Supersedes #683
Resolves #679
Fixes incorrect relogging logic, as pointed out in https://github.com/mojira/arisa-kt/pull/683#issuecomment-855244118

⚠️ These changes have not been tested yet, and are probably incomplete. Feel free to continue work here on this branch.
Due to this `catch` the relogging code will most likely not even run, so maybe we should remove that as well:
https://github.com/mojira/arisa-kt/blob/4ddec2350ba577fa8d1dccee393a0dc04e5ce43d/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt#L170-L175

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


violine1101:
In #689 I've replaced the `null` in the piece of code you quoted above by `throw e`, which _should_ propagate the exception and cause the non-successful branch to be run.

PR #691, opened 2021-06-09
Cache projects instead of re-querying them on every issue query

### LINKED ISSUE #688: Projects are loaded on each issue instead of being cached
## The Bug
Projects are loaded on each issue instead of being cached 
## Expected behavior
We cache it somewhere, as its currently 70% of our cpu usage

violine1101:
## Purpose
Fix #688 (Projects are loaded on each issue instead of being cached).

## Approach
Instead of querying the entire project every time, a cache will be used to store the projects for 5 minutes, after which the project is re-queried.

The cache can be force-cleared manually by using the `$ARISA_CLEAR_PROJECT_CACHE` command (e.g. in case a new snapshot comes out and Arisa considers it to still be a future version).

## Future work
It would be great to not be reliant on having the entire project info. Just updating the version list should be enough (similar to how MojiraBot does it). However, this is non-trivial with the JIRA library that we use.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/infrastructure/ProjectCache.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.github.mojira.arisa.jiraClient
+import java.time.Duration
+import java.time.Instant
+import net.rcarz.jiraclient.Project as JiraProject
+
+object ProjectCache {
+    private var cache = Cache<JiraProject>()
+    private var lastRefresh = Instant.now()
+
+    private const val REFRESH_INTERVAL_IN_MINUTES: Long = 5
```
urielsalis, line 12: We have the command to clear it and versions are created much earlier than they are actually released, how about we set it to an hour?

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/infrastructure/ProjectCache.kt
violine1101: The thing is, the version info includes stuff like "is this version released or not", which e.g. FutureModule relies on. With 5 minutes, the delay would be barely noticeable if nobody runs the command, while with 60 minutes effectively we're requiring someone to run the command during the first hour after version release.

I don't think the call should be _too_ expensive to do once every five minutes. Currently we're calling it every time a ticket is updated, so once every second or so.

If a call every five minutes is still too expensive, we can still up this number easiliy in a future PR.

src/main/kotlin/io/github/mojira/arisa/infrastructure/ProjectCache.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    fun getProjectFromTicketId(id: String): JiraProject {
+        checkForRefresh()
+
+        val projectKey = id.split("-")[0]
```
Marcono1234, line 17: This custom parsing logic might not be necessary, rcarz's `Issue` has a `getProject()` method from which we could get the project key.

Marcono1234 (commented):
Maybe we should actually have a version cache, and only update it when version information is actually requested. All other information of `Project` is immutable or does not need to be fetched. And there are only very few modules which actually need the version list and all of them are likely used pretty rarely: `FutureVersionModule` (only checks versions after having verified that issue has future version), and `AddVersionCommand` and `FixedCommand`.

However, I am not sure how relevant this is now since with your change projects are only fetched every 5 minutes anyways.

src/main/kotlin/io/github/mojira/arisa/infrastructure/ProjectCache.kt
violine1101: Yeah, I noticed this as well after this was merged... but I figured it doesn't really matter since in the end it does the same thing.

violine1101:
I was investigating whether we need to have the project objects themselves in the first place, and the answer is, we probably don't really. Most of the information can be gathered in other ways too. But I didn't really want to do larger architectural changes, especially since we plan to recode all of the rcarz ↔ Arisa domain conversion anyway.

When I submitted this PR I originally thought that the project of the issue would be dependant on the project cache, which would have had grave consequences, so I was very careful. Fortunately that turned out not to be the case.

PR #701, opened 2021-06-13
Refactor ArisaMain

violine1101:
## Purpose
`ArisaMain` got messier and messier, so it's time for some spring cleaning, innit?

## Approach
This turned out to be a bigger task than I expected.

This PR refactors `ArisaMain` by splitting it up into smaller pieces. Most notably:
* `JiraConnectionService` is responsible for connecting to Jira, and reconnecting if something goes wrong
* `HelperMessageUpdateService` is simply responsible to update the helper messages if they haven't been updated in a while. Simple.
* `LastRun` is responsible for keeping track of the previous run, saving it to the file, and reading it from the file when the bot starts initially.
* `ExecutionTimeframe` existed previously inside of `ModuleRegistry`, now it's its own class and handles everything in regard to the time frame in which the bot searches for updated tickets.
* `ArisaMain` is only responsible for glueing everything together.

I removed all of the timeouts / intervals from the config and into the kotlin classes themselves. I don't think it's necessary to have them in the config, and it's just easier to overview what something is set to by having it directly where it's used.

I've also fixed a small bug (?) where, if the bot would encounter a connection error, it would immediately go to sleep for 5 minutes, and then try to reconnect. This might not be optimal if there's only a very short outage. Instead, the bot now saves when the last successful connection happened, and only tries to relog if that connection is older than a minute.

Speaking of that, I didn't change any of the intervals from previously, only renamed them. I think they make sense how they were/are, but we can change them pretty easily if we want to.

As a summary, when Arisa encounters an exception it will do the following:
1. If the last successful connection (previously: last relog) was less than a minute ago: Wait 40 seconds
2. Try relogging
3. If that wasn't successful, wait 5 minutes and repeat from step 2
4. Continue as normal

Note, I've also removed the printing of the exception when a relog fails, in order to prevent spam. I'd say it suffices if we know that JIRA is not reachable, and at least one exception is always printed.

## Future work
MOAR REFACTORING

#### Checklist
- [ ] Included tests
- [x] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
@@ -1,165 +1,68 @@
 package io.github.mojira.arisa
 
-import arrow.syntax.function.partially1
 import ch.qos.logback.classic.AsyncAppender
 import ch.qos.logback.classic.LoggerContext
 import com.github.napstr.logback.DiscordAppender
 import com.uchuhimo.konf.Config
 import com.uchuhimo.konf.source.yaml
-import io.github.mojira.arisa.domain.Issue
-import io.github.mojira.arisa.infrastructure.HelperMessageService
-import io.github.mojira.arisa.infrastructure.ProjectCache
 import io.github.mojira.arisa.infrastructure.config.Arisa
-import io.github.mojira.arisa.infrastructure.jira.connectToJira
-import io.github.mojira.arisa.infrastructure.jira.toDomain
-import io.github.mojira.arisa.registry.TicketQueryTimeframe
-import io.github.mojira.arisa.registry.getModuleRegistries
 import net.rcarz.jiraclient.JiraClient
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import java.io.File
-import java.lang.Long.max
-import java.time.Duration
-import java.time.Instant
-import java.time.temporal.ChronoUnit
 import java.util.concurrent.TimeUnit
 
 val log: Logger = LoggerFactory.getLogger("Arisa")
 
-private const val TIME_MINUTES = 5L
-private const val MAX_QUERY_TIMEFRAME_IN_MINUTES = 10L
-
-const val MAX_RESULTS = 50
-
-private const val RELOG_INTERVAL_IN_SECONDS = 5 * 60L
-private const val WAIT_TIME_AFTER_CONNECTION_ERROR_IN_SECONDS = 40L
-
 lateinit var jiraClient: JiraClient
 
-@Suppress("LongMethod", "TooGenericExceptionCaught", "NestedBlockDepth")
+private val config = readConfig()
+private val connectionService = JiraConnectionService(config)
+private var executor = Executor(config)
+
+const val MIN_TIME_BETWEEN_EXECUTION_CYCLES_IN_SECONDS = 10L
+
 fun main() {
-    val config = readConfig()
-    setWebhookOfLogger(config)
+    setLoggerWebhooks()
 
-    jiraClient =
-        connectToJira(
-            config[Arisa.Credentials.username],
-            config[Arisa.Credentials.password],
-            config[Arisa.Issues.url]
-        )
-    log.info("Connected to jira")
-
-    // Get tickets for re-run and last run time
-    var lastRelog = Instant.now()
-    val lastRunFile = File("last-run")
-    val lastRun = readLastRun(lastRunFile)
-    var lastRunTime = readLastRunTime(lastRun)
-    var rerunTickets = lastRun.subList(1, lastRun.size).toSet()
-    val failedTickets = mutableSetOf<String>()
-
-    val checkIntervalSeconds = config[Arisa.Issues.checkIntervalSeconds]
-
-    // Read helper-messages
-    val helperMessagesFile = File("helper-messages.json")
-    val helperMessagesInterval = config[Arisa.HelperMessages.updateIntervalSeconds]
-    HelperMessageService.updateHelperMessages(helperMessagesFile)
-    var helperMessagesLastFetch = Instant.now()
-
-    // Create executor
-    val moduleRegistries = getModuleRegistries(config)
-    var executor = Executor(
-        config, moduleRegistries, getSearchIssues(jiraClient, config)
-    )
+    connectionService.connect()
 
     while (true) {
-        // Save time before run, so nothing happening during the run is missed
-        val currentTime = Instant.now()
-        val endOfMaxTimeframe = lastRunTime.plus(MAX_QUERY_TIMEFRAME_IN_MINUTES, ChronoUnit.MINUTES)
+        HelperMessageUpdateService.checkForUpdate()
 
-        // The time at which we execute the bot.
-        // If we exceed our maximum time frame, act as if we were executing at the end of the maximum time frame.
-        // This is to make sure that `last-run` is updated once in a while,
-        // even if the bot is catching up after a long downtime.
-        val runOpenEnded = currentTime.isBefore(endOfMaxTimeframe)
-        val currentRunTime = if (runOpenEnded) { currentTime } else { endOfMaxTimeframe }
-
-        val timeframe = TicketQueryTimeframe(lastRunTime, currentRunTime, runOpenEnded)
-
-        // Execute all enabled modules using the executor
-        val executionResults = executor.execute(timeframe, rerunTickets)
-
-        if (executionResults.successful) {
-            rerunTickets = emptySet()
-            failedTickets.addAll(executionResults.failedTickets)
-            val failed = failedTickets.joinToString("") { ",$it" } // even first entry should start with a comma
-
-            if (config[Arisa.Debug.updateLastRun]) {
-                lastRunFile.writeText("${currentRunTime.toEpochMilli()}$failed")
-            }
-            lastRunTime = currentRunTime
-            TimeUnit.SECONDS.sleep(checkIntervalSeconds)
-        } else {
-            if (Duration.between(lastRelog, Instant.now()).toMinutes() >= 1) {
-                // If last relog was more than a minute before and execution failed with an exception, relog
-                log.info("Trying to relog")
-                try {
-                    jiraClient =
-                        connectToJira(
-                            config[Arisa.Credentials.username],
-                            config[Arisa.Credentials.password],
-                            config[Arisa.Issues.url]
-                        )
-                    log.info("Relog was successful")
-                    lastRelog = Instant.now()
-                    executor = Executor(
-                        config, moduleRegistries, getSearchIssues(jiraClient, config)
-                    )
-                    TimeUnit.SECONDS.sleep(checkIntervalSeconds)
-                } catch (exception: Exception) {
-                    log.error("Relog failed", exception)
-                    // Wait before performing any further action
-                    TimeUnit.SECONDS.sleep(max(RELOG_INTERVAL_IN_SECONDS, checkIntervalSeconds))
-                }
-            } else {
-                // Not enough time for relog passed, just try waiting a bit
-                log.info("Not enough time for relog has passed")
-                TimeUnit.SECONDS.sleep(max(WAIT_TIME_AFTER_CONNECTION_ERROR_IN_SECONDS, checkIntervalSeconds))
-            }
-        }
-
-        // Update helper messages if necessary
-        if (currentTime.epochSecond - helperMessagesLastFetch.epochSecond >= helperMessagesInterval) {
-            HelperMessageService.updateHelperMessages(helperMessagesFile)
-            executor = Executor(
-                config, moduleRegistries, getSearchIssues(jiraClient, config)
-            )
-            helperMessagesLastFetch = currentTime
-        }
+        val secondsToSleep = runExecutionCycle(executor)
+        TimeUnit.SECONDS.sleep(secondsToSleep)
     }
 }
 
-private fun getSearchIssues(
-    jiraClient: JiraClient,
-    config: Config
-): (String, Int, () -> Unit) -> List<Issue> {
-    return ::searchIssues
-        .partially1(jiraClient)
-        .partially1(config)
+/**
+ * @return amount of seconds to sleep after this execution cycle
+ */
+private fun runExecutionCycle(executor: Executor): Long {
+    val timeframe = ExecutionTimeframe.getTimeframeFromLastRun()
+    val currentRunTime = timeframe.currentRunTime
+
+    // Execute all enabled modules using the executor
+    val executionResults = executor.execute(timeframe, LastRun.failedTickets)
+
+    return if (executionResults.successful) {
+        // Reset relog timer
+        connectionService.notifyOfSuccessfulConnection()
+
+        // Update last run and save it to file
+        LastRun.update(currentRunTime, executionResults.failedTickets)
+        if (config[Arisa.Debug.updateLastRun]) {
+            LastRun.writeToFile()
+        }
+
+        MIN_TIME_BETWEEN_EXECUTION_CYCLES_IN_SECONDS
+    } else {
+        connectionService.tryRelog().sleepTimeInSeconds
+    }
 }
 
-private fun readLastRunTime(lastRun: List<String>): Instant {
-    return if (lastRun[0].isNotEmpty())
-        Instant.ofEpochMilli(lastRun[0].toLong())
-    else Instant.now().minus(TIME_MINUTES, ChronoUnit.MINUTES)
-}
-
-private fun readLastRun(lastRunFile: File): List<String> {
-    return (if (lastRunFile.exists())
-        lastRunFile.readText().trim()
-    else "")
-        .split(",")
-}
-
+/**
+ * Reads the config from the config file(s) and other sources
+ */
 private fun readConfig(): Config {
     return Config { addSpec(Arisa) }
         .from.yaml.watchFile("config/config.yml")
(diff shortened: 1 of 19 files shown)


src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -1,165 +1,68 @@
...
+lateinit var jiraClient: JiraClient
 
-private const val RELOG_INTERVAL_IN_SECONDS = 5 * 60L
-private const val WAIT_TIME_AFTER_CONNECTION_ERROR_IN_SECONDS = 40L
+private val config = readConfig()
+private val connectionService = JiraConnectionService(config)
+private var executor = Executor(config)
 
-lateinit var jiraClient: JiraClient
+const val MIN_TIME_BETWEEN_EXECUTION_CYCLES_IN_SECONDS = 10L
```
urielsalis, line 22: We can use 10.seconds here (yay for kotlin)

I think its out of experimental

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -1,165 +1,68 @@
...
+        MIN_TIME_BETWEEN_EXECUTION_CYCLES_IN_SECONDS
+    } else {
+        connectionService.tryRelog().sleepTimeInSeconds
+    }
 }
 
+/**
+ * Reads the config from the config file(s) and other sources
+ */
 private fun readConfig(): Config {
```
urielsalis, line 66: I would move all the config stuff to its own service too

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -168,46 +71,22 @@ private fun readConfig(): Config {
         .from.systemProperties()
 }
 
-private fun setWebhookOfLogger(config: Config) {
+/**
+ * Configures the logger, so that it sends log messages to the Discord webhooks
+ */
+private fun setLoggerWebhooks() {
```
urielsalis, line 77: and the webhooks too 👀

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -1,165 +1,68 @@
...
-): (String, Int, () -> Unit) -> List<Issue> {
-    return ::searchIssues
-        .partially1(jiraClient)
-        .partially1(config)
-}
+/**
+ * @return amount of seconds to sleep after this execution cycle
+ */
+private fun runExecutionCycle(executor: Executor): Long {
+    val timeframe = ExecutionTimeframe.getTimeframeFromLastRun()
```
urielsalis, line 41: maybe this whole loop can be a ExecutionService and Arisa just initializes it and calls a run on it?

urielsalis: The idea is that the main is as small as possible

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
violine1101: Hmm, do we need to? It's literally only this function.

src/test/kotlin/io/github/mojira/arisa/ExecutorTest.kt
```diff
@@ -24,7 +23,7 @@ private val failedModuleRegistryMock = mockk<ModuleRegistry>()
 private val moduleExecutorMock = mockk<ModuleExecutor>()
 private val failedModuleExecutorMock = mockk<ModuleExecutor>()
 
-private val dummyTimeframe = TicketQueryTimeframe(Instant.now(), Instant.now(), true)
+private val dummyTimeframe = ExecutionTimeframe(Instant.now(), Instant.now(), true)
```
urielsalis, line 26: Would be nice to see if we can add tests for all the new classes now that they are separate, that would avoid most of the errors we are having lately

urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
urielsalis: Ideally Main doesnt even need the config, so this can be its own service and we call it when we need it(multiple instances of it is fine)

src/test/kotlin/io/github/mojira/arisa/ExecutorTest.kt
violine1101: Yeah, adding tests for all these classes would be pretty important. I'm not sure if I'll have enough time this week to add some for this. If not we should definitely keep it in mind.

FWIW, I've tested this all locally in a dry run and it seems to work exactly as before, so it should be fine?

At the very least I didn't have to remove any tests this time 😅

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
violine1101: I tried extracting this into another file. The problem is, this function right here relies on quite a lot of stuff in `ArisaMain`, like `config`, `connectionService`, and `executor`. I didn't think it was worth it, and having all the "glue" here in one place is pretty good imo.

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
violine1101: Yeah, that works. Might be nice to change it in the future, but that'd be another larger change since that'd cascade down to other things too I think.

Marcono1234: Looks like it is still experimental, see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/
Though it would indeed be good to refactor this and other time related checks in the future. Personally I find these `plus(x).isBefore(...)` checks more difficult to understand than using JDK `Duration`, e.g. `Duration.between(...).toMinutes() < ...`.

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
```diff
@@ -0,0 +1,134 @@
...
+         */
+        private const val WAIT_TIME_AFTER_UNSUCCESSFUL_RELOG_IN_SECONDS = 5 * 60L // 5 minutes
+
+        /**
+         * How long Arisa should wait after a connection error before continuing without trying to relog
+         */
+        private const val WAIT_TIME_AFTER_CONNECTION_ERROR_IN_SECONDS = 40L
+    }
+
+    private var lastSuccessfulConnection = Instant.now().minusSeconds(MAX_SECONDS_SINCE_LAST_SUCCESSFUL_CONNECTION + 1)
```
Marcono1234, line 31: What is the rationale for this value? Relog immediately if initial connection attempt fails?
Might be good to add a comment explaining the value.

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
```diff
@@ -0,0 +1,134 @@
...
+     * - [RelogResult.UnsucessfulRelog] if unable to relog
+     * - [RelogResult.NoRelogAttempted] if no relog was attempted because the last relog happened recently
+     */
+    fun tryRelog(): RelogResult {
+        val secondsSinceLastSuccessfulConnection = Duration.between(lastSuccessfulConnection, Instant.now()).toSeconds()
+
+        return if (secondsSinceLastSuccessfulConnection > MAX_SECONDS_SINCE_LAST_SUCCESSFUL_CONNECTION) {
+            log.info("Trying to relog")
+
+            establishConnection() ?: return RelogResult.SuccessfulRelog()
```
Marcono1234, line 69: Might be good to at least log the exception at DEBUG level.

Marcono1234, line 69: For sanity should probably call `notifyOfSuccessfulConnection()` after successful relog. Otherwise if for some other reason execution keeps failing (e.g. broken module) Arisa would continuously relog (with only `MIN_TIME_BETWEEN_EXECUTION_CYCLES_IN_SECONDS` delay in between).

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
```diff
@@ -0,0 +1,134 @@
...
+
+            val relogResult = RelogResult.UnsucessfulRelog()
+            log.error(
+                "Could not reconnect. Will attempt to relog again in ${ relogResult.sleepTimeInSeconds } seconds."
+            )
+
+            relogResult
+        } else {
+            RelogResult.NoRelogAttempted()
```
Marcono1234, line 78: Might be good to log an INFO message so we understand why Arisa does not appear to be doing anything when it waits after no attempted relog.

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
violine1101: Hm, personally I have troubles with `Duration.between` because it might be negative and do unexpected things... so just to be sure I always put `.abs()` there which is annoying and I don't like it. Not sure if there's a better option.

violine1101: Ah, also it seems like what I experimented with was arrow's version of the `.seconds` suffix, so that was probably something different than you meant.

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
violine1101: Hmm, I'm not too sure myself. I think it's a remnant of debugging stuff, I was running into infinite loops when starting the bot and connection failed and this solved it. But I think the actual bug was elsewhere.

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -1,165 +1,68 @@
...
-private const val MAX_QUERY_TIMEFRAME_IN_MINUTES = 10L
-
-const val MAX_RESULTS = 50
+lateinit var jiraClient: JiraClient
 
-private const val RELOG_INTERVAL_IN_SECONDS = 5 * 60L
-private const val WAIT_TIME_AFTER_CONNECTION_ERROR_IN_SECONDS = 40L
+private val config = readConfig()
+private val connectionService = JiraConnectionService(config)
+private var executor = Executor(config)
```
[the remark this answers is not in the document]
violine1101: Yes it can! Previously the executor was constantly changed from the outside, I've changed it so that it now always accesses the global `jiraClient` instead of having a separate copy (which is not very nice either but it works and we don't need to allocate a new executor on every relog).

src/main/kotlin/io/github/mojira/arisa/LastRun.kt
```diff
@@ -0,0 +1,42 @@
+package io.github.mojira.arisa
+
+import java.io.File
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+/**
+ * Stores information about the previous run (start time, and tickets that failed during the run)
+ */
+object LastRun {
```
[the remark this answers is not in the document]
violine1101: Yeah... personally I'm kinda fond of the `objects`, they're kinda nice, but I feel like it's dangerous to fall into the "make everything global" antipattern. Honestly didn't think about testability. Moving away from objects would probably be a better idea in the long run.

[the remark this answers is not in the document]
violine1101: I'm holding off on changing `HelperMessageUpdateService` and `HelperMessageService` for now, that would involve some larger changes to the rest of the codebase. Something to keep in mind though.

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
violine1101: Yeah, I wish we wouldn't need `notifyOfSuccessfulConnection`, it seems like it's risky that we might forget it somewhere. And of course I did.

src/main/kotlin/io/github/mojira/arisa/JiraConnectionService.kt
violine1101: Simply using `Instant.now()` seems to work just fine, not sure why I did it this way

violine1101:
Current status:

* This is basically done, I don't plan to make any further changes to Arisa's code itself.
* Not sure why the build still fails. It passes locally.
* I'd really like to add tests for this, given that this is a large change and this area of the code is insufficiently tested as is. However that turned out to be rather difficult to do and I don't have the time to look into that further at the moment.

src/main/kotlin/io/github/mojira/arisa/ArisaMain.kt
```diff
@@ -1,213 +1,32 @@
...
-    var lastRelog = Instant.now()
-    val lastRunFile = File("last-run")
-    val lastRun = readLastRun(lastRunFile)
-    var lastRunTime = readLastRunTime(lastRun)
-    var rerunTickets = lastRun.subList(1, lastRun.size).toSet()
-    val failedTickets = mutableSetOf<String>()
+val configService = ConfigService()
+val webhookService = WebhookService(configService.config)
+val connectionService = JiraConnectionService(configService.config)
+val executionService = ExecutionService(configService.config, connectionService)
```
Marcono1234, line 21: These could be `private`, or even better local variables of `main`; this would then most likely solve the build failure.

Marcono1234 (commented):
> * Not sure why the build still fails. It passes locally.

The underlying issue appears to be this:
```
 io.github.mojira.arisa.ExecutorTest > should add failed tickets FAILED
    java.lang.ExceptionInInitializerError
        at io.github.mojira.arisa.Executor.<init>(Executor.kt:103)
        at io.github.mojira.arisa.ExecutorTestKt.getMockExecutor(ExecutorTest.kt:93)
        at io.github.mojira.arisa.ExecutorTestKt.getMockExecutor$default(ExecutorTest.kt:89)
        at io.github.mojira.arisa.ExecutorTest$1$10.invokeSuspend(ExecutorTest.kt:55)
        at io.github.mojira.arisa.ExecutorTest$1$10.invoke(ExecutorTest.kt)

        Caused by:
        java.io.FileNotFoundException: /home/runner/work/arisa-kt/arisa-kt/config/local.yml (No such file or directory)
            at java.base/java.io.FileInputStream.open0(Native Method)
            at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
            at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
            at com.uchuhimo.konf.source.Provider$DefaultImpls.file(Provider.kt:83)
            at com.uchuhimo.konf.source.yaml.YamlProvider.file(YamlProvider.kt:37)
            at com.uchuhimo.konf.source.Loader.watchFile(Loader.kt:119)
            at com.uchuhimo.konf.source.Loader.watchFile(Loader.kt:194)
            at com.uchuhimo.konf.source.Loader.watchFile$default(Loader.kt:192)
            at io.github.mojira.arisa.ConfigService.<init>(ConfigService.kt:10)
            at io.github.mojira.arisa.ArisaMainKt.<clinit>(ArisaMain.kt:18)
            ... 5 more
```
(see https://github.com/mojira/arisa-kt/runs/2944775815)

This seems to be caused by `Executor`'s reference to `ArisaMainKt`'s `log`. This then initializes the class and tries to load the config (other test classes seem to have a similar problem).

In general avoiding usage of one logger in `ArisaMainKt` for multiple classes might be a good idea to prevent cyclic dependencies, though for a logger that might not matter that much.

violine1101:
Ah, yeah, thank you. That explains why it didn't happen locally too.

Regarding the global logger, it's definitely possible to no longer rely on that and only have a single logger in the executor, however we'd probably need some larger refactoring for that change to make sense.

Marcono1234:
Do you want to update the pull request with the changes from `master` and then mark it as "ready for review"?
I would review it then again.

violine1101:
Yeah... really I'd like to have these changes fully tested since they're quite extensive, but I have neither the time nor the motivation at the moment for that, so I guess an thorough review needs to suffice for now unless someone else wants to take care of writing the tests first.

src/test/kotlin/io/github/mojira/arisa/ExecutionTimeframeTest.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val lastRun = LastRun(
+            { lastRunTime.toEpochMilli().toString() },
+            { }
+        )
+
+        val timeframe = ExecutionTimeframe.getTimeframeFromLastRun(lastRun)
+
+        timeframe.lastRunTime shouldBe lastRunTime
+        timeframe.currentRunTime.isAfter(timeframeEnd) shouldBe false
```
Marcono1234, line 49: Maybe should test `isBefore` to make sure `currentRunTime != timeframeEnd`.

src/test/kotlin/io/github/mojira/arisa/LastRunTest.kt
```diff
@@ -0,0 +1,75 @@
...
+class LastRunTest : StringSpec({
+    "should use default time before now if input file is empty" {
+        val lastRun = LastRun(
+            { "" },
+            { }
+        )
+
+        val time = Instant.now().minus(LastRun.DEFAULT_START_TIME_MINUTES_BEFORE_NOW, ChronoUnit.MINUTES)
+
+        lastRun.time.isAfter(time) shouldBe false
```
Marcono1234, line 19: Since this test is already using `LastRun.DEFAULT_START_TIME_MINUTES_BEFORE_NOW`, it could possibly verify that `lastRun.time` is roughly `time` (with a leeway of for example 30 seconds).

Marcono1234 (approved):
Thanks a lot! The changes look good, from my side this is ready to be merged. @urielsalis, what do you think?

I have left a few review comments, but they are not really important and definitely not blocking integration of these changes.

PR #709, opened 2021-07-02
Redact access tokens in attachments

### LINKED ISSUE #704: Auto-redact JVM crash reports
## The Problem
JVM crash reports contain privacy sensitive information that causes any bug report with one to be privated immediately.

## Possible Solution
Automate what I've done in https://bugs.mojang.com/browse/MC-229049:
* Parse `hs_err_pid` crashes
* Regex privacy sensitive parts and replace them with `<redacted>`
* Attach redacted crash log with name `redacted_hs_err_pidXXXX.log`
* Delete original crash log

Marcono1234:
## Purpose
Fixes #704

Redact access tokens in attachments (e.g. JVM crash reports) by deleting the original attachment and uploading a redacted version to avoid having to make the issue private. Additionally add a comment to let the user know that their attachment has been redacted.

Open questions:
- [x] Should a (bot/hidden) helper message be used for the comment? If so it would require at least two placeholders: Uploader, new attachment names
Edit: Message is currently hardcoded since only one placeholder is supported at the moment, and message is only relevant for Arisa, but is not useful as helper message.

Open tasks:
- [x] Verify that it works with an actual JVM crash report
- [x] Update documentation
- [x] Test in MCTEST

## Approach
`PrivacyModule` has been adjusted to redact access token references in the form `--accessToken ...` like they appear in JVM crash reports. In theory this could be done in a separate module, though it is closely related to what `PrivacyModule` does and having a separate module do this would be tricky (see #527) and would require that the modules are executed in the correct order.

Redaction logic does not reuse `PrivacyModule` regex patterns because they are known to cause false positives and might also not cover everything, e.g. `braintree:` transaction ID is usually provided together with a order ID. We can reliably match the transaction ID (due to its unambiguous `braintree:` prefix), but matching the order ID is not possible. So in that case merely redacting the transaction ID would not suffice.
Therefore redaction only covers the access token for now.

As a safe guard the redacted attachment content is checked to see if it still contains sensitive data (since redaction does not cover everthing); if this is the case the attachment is not redacted and the issue is made private.

A publicly visible comment is added informing the user that their attachments have been redacted.

## Future work
An error during deletion of the existing attachment or upload of the redacted attachment might not be handled correctly; this can lead to the following unwanted situations:
- deletion of old attachment fails, report is made public even though it still contains sensitive attachment
- upload of new attachment fails, attachment is effectively lost

Maybe in the future we could change `IssueUpdateContext` to group actions done by a single module invocation, and abort execution for that module if one of its actions fails (similar to #720); that should solve the above situations. It would also require removing the logging from the `deleteAttachment` and `addAttachmentFile` helper functions and always throw an exception when the HTTP code indicates a failure, because we cannot be sure if we interpret the error HTTP code the way Jira meant it.

#### Checklist
- [x] Included tests
- [x] Tested in MCTEST-142

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
@@ -1,6 +1,7 @@
 package io.github.mojira.arisa.infrastructure
 
 import arrow.core.Either
+import arrow.core.getOrElse
 import arrow.core.left
 import arrow.core.right
 import arrow.core.rightIfNotNull
@@ -119,11 +120,18 @@ object HelperMessageService {
         }
     }
 
+    private const val BOT_SIGNATURE_KEY = "i-am-a-bot"
     fun getMessageWithBotSignature(project: String, key: String, filledText: String? = null, lang: String = "en") =
-        getMessage(project, listOf(key, "i-am-a-bot"), listOf(filledText), lang)
+        getMessage(project, listOf(key, BOT_SIGNATURE_KEY), listOf(filledText), lang)
 
     fun getMessageWithDupeBotSignature(project: String, key: String, filledText: String? = null, lang: String = "en") =
         getMessage(project, listOf(key, "i-am-a-bot-dupe"), listOf(filledText), lang)
+
+    fun getRawMessageWithBotSignature(rawMessage: String): String {
+        // Note: Project does not matter, message is (currently) the same for all projects
+        val botSignature = getSingleMessage("MC", BOT_SIGNATURE_KEY).getOrElse { "" }
+        return "$rawMessage\n$botSignature"
+    }
 
     fun setHelperMessages(json: String) = data.fromJSON(json)
         ?: throw IOException("Couldn't deserialize helper messages from setHelperMessages()")
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
@@ -52,7 +52,7 @@ fun JiraAttachment.toDomain(jiraClient: JiraClient, issue: JiraIssue, config: Co
 )
 
 fun getCreationDate(issue: JiraIssue, id: String, default: Instant) = issue.changeLog.entries
-    .filter { it.items.any { it.field == "Attachment" && it.to == id } }
+    .filter { changeLogEntry -> changeLogEntry.items.any { it.field == "Attachment" && it.to == id } }
     .maxByOrNull { it.created }
     ?.created
     ?.toInstant() ?: default
@@ -221,11 +221,12 @@ fun JiraComment.toDomain(
 fun JiraUser.toDomain(jiraClient: JiraClient, config: Config) = User(
     name, displayName,
     ::getUserGroups.partially1(jiraClient).partially1(name),
-    ::isNewUser.partially1(jiraClient).partially1(name),
+    ::isNewUser.partially1(jiraClient).partially1(name)
+) {
     // Check case insensitively because it apparently does not matter when logging in, so `username` might have
     // incorrect capitalization
-    { name.equals(config[Arisa.Credentials.username], ignoreCase = true) }
-)
+    name.equals(config[Arisa.Credentials.username], ignoreCase = true)
+}
 
 private fun getUserGroups(jiraClient: JiraClient, username: String) = getGroups(
     jiraClient,
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
@@ -204,7 +204,9 @@ fun JiraUser.toDomain(jiraClient: JiraClient, config: Config) = User(
     name, displayName,
     ::getUserGroups.partially1(jiraClient).partially1(name),
     ::isNewUser.partially1(jiraClient).partially1(name),
-    { name == config[Arisa.Credentials.username] }
+    // Check case insensitively because it apparently does not matter when logging in, so `username` might have
+    // incorrect capitalization
+    { name.equals(config[Arisa.Credentials.username], ignoreCase = true) }
 )
 
 private fun getUserGroups(jiraClient: JiraClient, username: String) = getGroups(
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
@@ -37,15 +37,16 @@ import net.rcarz.jiraclient.Project as JiraProject
 import net.rcarz.jiraclient.User as JiraUser
 import net.rcarz.jiraclient.Version as JiraVersion
 
-fun JiraAttachment.toDomain(jiraClient: JiraClient, issue: JiraIssue) = Attachment(
+fun JiraAttachment.toDomain(jiraClient: JiraClient, issue: JiraIssue, config: Config) = Attachment(
     id,
     fileName,
     getCreationDate(issue, id, issue.createdDate.toInstant()),
     mimeType,
     ::deleteAttachment.partially1(issue.getUpdateContext(jiraClient)).partially1(this),
     { openAttachmentStream(jiraClient, this) },
-    this::download,
-    author?.toDomain(jiraClient)
+    // Cache attachment content once it has been downloaded
+    lazy { this.download() }::value,
+    author?.toDomain(jiraClient, config)
 )
 
 fun getCreationDate(issue: JiraIssue, id: String, default: Instant) = issue.changeLog.entries
@@ -90,7 +91,7 @@ fun JiraIssue.toDomain(
         description,
         getEnvironment(),
         security?.id,
-        reporter?.toDomain(jiraClient),
+        reporter?.toDomain(jiraClient, config),
         resolution?.name,
         createdDate.toInstant(),
         updatedDate.toInstant(),
@@ -105,10 +106,10 @@ fun JiraIssue.toDomain(
         getDungeonsPlatform(config),
         mapVersions(),
         mapFixVersions(),
-        mapAttachments(jiraClient),
-        mapComments(jiraClient),
+        mapAttachments(jiraClient, config),
+        mapComments(jiraClient, config),
         mapLinks(jiraClient, config),
-        getChangeLogEntries(jiraClient),
+        getChangeLogEntries(jiraClient, config),
         ::reopen.partially1(context),
         ::resolveAs.partially1(context).partially1("Awaiting Response"),
         ::resolveAs.partially1(context).partially1("Invalid"),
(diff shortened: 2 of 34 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/HelperMessages.kt
```diff
@@ -119,12 +120,19 @@ object HelperMessageService {
...
     fun getMessageWithBotSignature(project: String, key: String, filledText: String? = null, lang: String = "en") =
-        getMessage(project, listOf(key, "i-am-a-bot"), listOf(filledText), lang)
+        getMessage(project, listOf(key, BOT_SIGNATURE_KEY), listOf(filledText), lang)
 
     fun getMessageWithDupeBotSignature(project: String, key: String, filledText: String? = null, lang: String = "en") =
         getMessage(project, listOf(key, "i-am-a-bot-dupe"), listOf(filledText), lang)
 
+    fun getRawMessageWithBotSignature(rawMessage: String): String {
+        // Note: Project does not matter, message is (currently) the same for all projects
+        val botSignature = getSingleMessage("MC", BOT_SIGNATURE_KEY).getOrElse { "" }
```
Marcono1234, line 132: Hardcoding the project key here might be a little bit hacky. What do you think?

src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt
```diff
@@ -28,67 +34,202 @@ class PrivacyModule(
...
+                        // Use link for attachments
+                        "\n- [^${sanitizeCommentArg(it)}]"
+                    }
+                    val sanitizedUploaderName = sanitizeCommentArg(uploader)
+                    // Does not use helper message because message will only be used by bot and helper messages
+                    // currently only support one placeholder
+                    issue.addRawBotComment(
+                        "@[~$sanitizedUploaderName], sensitive data has been removed from your " +
+                            "attachment(s) and they have been re-uploaded as:$fileNamesString"
+                    )
```
Marcono1234, line 229: Message is hardcoded instead of using helper message because:
- It uses two placeholders but helper messages only support one at the moment
- This message is only useful for Arisa, as helper message it does not add that much value since the user would have to format the `$fileNamesString` filltext manually.

Marcono1234:
@violine1101, since you originally proposed this feature, could you please have a look when you have some spare time?

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
```diff
@@ -0,0 +1,242 @@
...
+                        // Use link for attachments
+                        "\n- [^${sanitizeCommentArg(it)}]"
+                    }
+                    val sanitizedUploaderName = sanitizeCommentArg(uploader)
+                    // Does not use helper message because message will only be used by bot and helper messages
+                    // currently only support one placeholder
+                    issue.addRawBotComment(
+                        "@[~$sanitizedUploaderName], sensitive data has been removed from your " +
+                            "attachment(s) and they have been re-uploaded as:$fileNamesString"
+                    )
```
Marcono1234, line 236: Message is hardcoded instead of using helper message because:
- It uses two placeholders but helper messages only support one at the moment
- This message is only useful for Arisa, as helper message it does not add that much value since the user would have to format the `$fileNamesString` filltext manually.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
```diff
@@ -0,0 +1,242 @@
...
+                        // Remove the original attachment
+                        attachment.remove()
+                    }
+                }
+
+                if (fileNames.isNotEmpty()) {
+                    // Use postfix line break to prevent bot signature from appearing as part of last list item
+                    val fileNamesString = fileNames.joinToString(separator = "", postfix = "\n") {
+                        // Use link for attachments
+                        "\n- [^${sanitizeCommentArg(it)}]"
```
violine1101, line 228: The aggressive sanitizing here might lead to some attachment links not working, in case they include unicode characters. Which will, arguably, probably barely ever happen. So not sure if it's a big deal.

violine1101: I guess we run the file names through a regex firstly though, so this isn't an issue.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private data class AttachmentsCheckResult(
+        /** Whether any of the attachments contains sensitive data which cannot be redacted */
+        val attachmentContainsSensitiveData: Boolean,
+        val attachmentsToRedact: List<RedactedAttachment>
+    )
+
+    private fun Issue.checkAttachments(lastRun: Instant): AttachmentsCheckResult {
+        var attachmentContainsSensitiveData: Boolean
```
violine1101, line 86: I feel like the name of this variable is not very good, I got quite confused reading this.

violine1101: Maybe something like `sensitiveDataFound`?

I guess the thing that confused me most is that this variable is not about a particular attachment, but rather about if _any_ attachment contains _any_ sensitive data.

violine1101 (commented):
This looks great! Thanks for working on this!

Apart from a few comments, most of which aren't really actionable, I think this should be good to go.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
```diff
@@ -0,0 +1,242 @@
...
+                assertNotEmpty(attachmentsToRedact),
+                assertTrue(issueContainsSensitiveData),
+                assertNotEmpty(restrictCommentActions)
+            ).bind()
+
+            // Always try to redact attachments, even if issue would be made private anyways
+            // So in case issue was made private erroneously it can easily be made public
+            val redactedAll = redactAttachments(issue, attachmentsToRedact)
+
+            if (!redactedAll || issueContainsSensitiveData) {
```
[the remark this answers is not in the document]
violine1101: I guess I got confused by the variable naming... it's talking about all attachments containing sensitive data, not all attachments of the ticket.

I guess a more proper variable naming might be `redactingFailed` or something similar (for the negation).

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
```diff
@@ -39,19 +39,20 @@ import net.rcarz.jiraclient.Project as JiraProject
 import net.rcarz.jiraclient.User as JiraUser
 import net.rcarz.jiraclient.Version as JiraVersion
 
-fun JiraAttachment.toDomain(jiraClient: JiraClient, issue: JiraIssue) = Attachment(
+fun JiraAttachment.toDomain(jiraClient: JiraClient, issue: JiraIssue, config: Config) = Attachment(
```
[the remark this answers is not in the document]
Marcono1234: Maybe we could also in the future remove more hardcoded values (e.g. Mojira user group names) and add them to the config. Then we could benefit from the config being available here as well.
But you are right, it is a bit annoying that the config has to be passed around to so much functions.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
Marcono1234: We can make `sanitizeCommentArg` more lenient (e.g. by allowing certain character categories) in the future if this turns out to be a problem.

> I guess we run the file names through a regex firstly though

Which one do you mean? We only have a list of regex patterns for names of files which definitely contain sensitive data, but we do not filter malformed file names (except to prevent directory traversal).

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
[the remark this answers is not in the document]
urielsalis: Something I was thinking is saving the config to a public car somewhere and just calling that statically instead of passing it around

Less functional, but prettier

urielsalis (dismissed):

src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mappers.kt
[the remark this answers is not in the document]
Marcono1234: We could also make `Mappers` a class which has `JiraClient` and `Config` as properties, similar to your [`MapFromJira`](https://github.com/mojira/arisa-kt/blob/9a354f3c2e264cc7eb01f3fcb4174921b0ac5b6e/src/main/kotlin/io/github/mojira/arisa/infrastructure/newjira/MapFromJira.kt) class from #720, or my (incomplete) [changes on the `marcono1234/refactor` branch](https://github.com/mojira/arisa-kt/blob/marcono1234/refactor/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Mapper.kt).
This can greatly reduce code duplication.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
Marcono1234: Changed it to `foundNonRedactableSensitiveData`; a bit verbose but hopefully describes its purpose well enough.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
```diff
@@ -99,37 +99,37 @@ class PrivacyModule(
...
-                    if (!attachmentContainsSensitiveData) {
+                    if (!foundNonRedactableSensitiveData) {
                         containsSensitiveData(it.getTextContent())?.let { matchResult ->
-                            attachmentContainsSensitiveData = true
+                            foundNonRedactableSensitiveData = true
                             logFoundSensitiveData("in attachment with ID ${it.id}", matchResult)
                         }
                     }
                     return@mapNotNull null
-                } else {
```
Marcono1234: Changed this from:
```
else {
  if (...) {
    return ...
  }
  return
}
```
to
```
else if (...) {
  return ...
} else {
  return
}
```

These are (except for the variable name change) only indentation changes.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
violine1101: From what I can tell, attachments are only scanned if they match the hs_err_pid regex. I might be wrong though.

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
violine1101: As this is Kotlin and if statements are expressions, in theory you should be able to put the `return` in front of the `if` too, I believe.

Marcono1234: In theory the `return@mapNotNull` could be removed as well, but at least at the moment I am not convinced by that Kotlin feature yet, at least for larger code blocks as it is the case here. In my opinion that looks too similar to an unused expression (which Kotlin allows, but at least IntelliJ warns about it).

src/main/kotlin/io/github/mojira/arisa/modules/privacy/PrivacyModule.kt
Marcono1234: It has been some time since I wrote that code, but I don't think there are such requirements. `AccessTokenRedactor` only checks for the file content (more specifically for `--accessToken`), and the `PrivacyModuleTest` passes without specifying any special file names.

violine1101: Ah yes you're right. That check exists but it doesn't affect the redacting.

violine1101 (approved):

Marcono1234:
Regarding the "Future work" mentioned in the description: Implementing this might be tricky / not easily possible with the way Jira actions are currently performed. Because `IssueUpdateContext` groups actions by how they are performed against Jira, performing them in the order in which a module executed them is probably difficult without sacrificing efficient grouping of the actions.

PR #720, opened 2021-07-11
Domain, Jira, Major refactor: Adds the new domain classes and mapping from Jira

urielsalis:
## Purpose
Adds the new domain classes and mapping from Jira so we can move to a model where we apply imperative shell functional core
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


src/main/kotlin/io/github/mojira/arisa/newdomain/Attachment.kt
```diff
@@ -0,0 +1,15 @@
...
+
+data class Attachment(
+    val id: String,
+    val name: String,
+    val created: Instant,
+    val mimeType: String,
+    val openContentStream: () -> InputStream,
+    val getContent: () -> ByteArray,
+    val uploader: User?,
+    var exists: Boolean = true
```
Marcono1234, line 14: What is this property for?

src/main/kotlin/io/github/mojira/arisa/infrastructure/newjira/MapFromJira.kt
```diff
@@ -0,0 +1,124 @@
...
+        this.resolution?.name,
+        this.createdDate.toInstant(),
+        this.updatedDate?.toInstant(),
+        this.resolutionDate?.toInstant(),
+        this.getFieldAsString(config[Arisa.CustomFields.chkField]),
+        this.getCustomField(config[Arisa.CustomFields.confirmationField]),
+        this.getFieldAsDouble(config[Arisa.CustomFields.linked]),
+        this.getCustomField(config[Arisa.CustomFields.mojangPriorityField]),
+        this.getFieldAsString(config[Arisa.CustomFields.triagedTimeField]),
+        this.project.toDomain(),
```
Marcono1234, line 43: Not sure if that would work, if I recall correctly it would not have the versions list. We would probably have to use `ProjectCache`.

Marcono1234 (commented):
If I understand it correctly, the "imperative shell, functional core" principle uses immutable data classes, but you are using mutable collections or `var` mutable properties. How would modification of an issue or similar work? Would a module simply modify these properties?

If so, then that might be a bit error prone because there would not be any "handler" functions anymore, but the executor would have to check which properties were modified. This might also make testing more verbose than our current solution where we can simply mock the handler functions.

Or how do you intend usage of these classes?

urielsalis:
> If I understand it correctly, the "imperative shell, functional core" principle uses immutable data classes, but you are using mutable collections or `var` mutable properties. How would modification of an issue or similar work? Would a module simply modify these properties?
> 
> If so, then that might be a bit error prone because there would not be any "handler" functions anymore, but the executor would have to check which properties were modified. This might also make testing more verbose than our current solution where we can simply mock the handler functions.
> 
> Or how do you intend usage of these classes?

Yes, the main idea is a funtional immutable core

With the refactor our idea is modifying the domain classes directly with the modules then having a layer above persist the changes(by comparing with the original issue), so it goes a bit in another direction, but what I wanted to maintain is that wour communication with jira is a small layer on top of everything that just translates tickets to domain and domain back to tickets, and all of our logic is completly independant

src/main/kotlin/io/github/mojira/arisa/newdomain/Attachment.kt
urielsalis: set it to false = delete attachment

PR #721, opened 2021-07-13
Make sure fixed command version was released after all affected versions

### LINKED ISSUE #696: `FixedCommand` should reject versions older than affected versions
## The Problem
(This has been brought up by @NeunEinser internally)

The issue is that game versions are pretty similar and it is easy to mistype them, e.g. writing `20w17a` instead of `21w17a` by accident. Since archived fixed versions cannot be removed again it is impossible to correct this mistake afterwards.

## Possible Solution
`FixedCommand` should reject versions which are older than the earliest affected version.
If for some reason the affected versions are incomplete, the user should manually add them using `AddVersionCommand` and afterwards set the fixed version.

Note that `AddVersionCommand` should probably not be adjusted, there might be cases where a report does not include the earliest version in which a bug occurred, so it might be desired to add an earlier affected version.

chandler05:
## Purpose
Fix #696
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
@@ -14,7 +14,7 @@ class FixedCommand {
         if (issue.resolution !in listOf(null, "", "Unresolved")) {
             throw CommandExceptions.ALREADY_RESOLVED.create(issue.resolution)
         }
-        if (issue.affectedVersions.any { it.releaseDate.isAfter(issue.project.versions.first { v -> v.name == version }.releaseDate) }) {
+        if (issue.affectedVersions.any { it.releaseDate!!.isAfter(issue.project.versions.first { v -> v.name == version }.releaseDate) }) {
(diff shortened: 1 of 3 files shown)


urielsalis (approved):

src/main/kotlin/io/github/mojira/arisa/modules/commands/FixedCommand.kt
```diff
@@ -14,6 +14,14 @@ class FixedCommand {
         if (issue.resolution !in listOf(null, "", "Unresolved")) {
             throw CommandExceptions.ALREADY_RESOLVED.create(issue.resolution)
         }
+        if (issue.affectedVersions.any {
+                it.releaseDate!!.isAfter(issue.project.versions.first {
+                        v -> v.name == version
+                }.releaseDate)
```
Marcono1234, line 20: Might be good to make this more efficient by avoiding the repeated lookup of
```kotlin
issue.project.versions.first { v -> v.name == version }.releaseDate
```
and only calculating it once.

PR #722, opened 2021-07-13
Only check first affected version in fixed command

chandler05:
## Purpose
https://github.com/mojira/arisa-kt/issues/696#issuecomment-879188249
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


urielsalis (approved):

Marcono1234:
@chandler05, sorry if https://github.com/mojira/arisa-kt/issues/696#issuecomment-879188249 sounded rude. Thank you for resolving the issue in the first place! Your idea of failing if any affected version is newer than the fix version is actually pretty good. I have tried re-implementing that in #728, but allowing to skip the check to forcefully set a fix version.

PR #734, opened 2021-07-24
Support custom messages for resolving issue with future version as Invalid

Marcono1234:
## Purpose
Follow-up for #725
Resolves https://github.com/mojira/helper-messages/issues/161

See also https://github.com/mojira/helper-messages/issues/161#issuecomment-881977354

Instead of resolving all reports with only a future version as Invalid, specify a mapping from version Jira ID to custom message key for resolving reports as Invalid. All other reports will still be resolved as Awaiting Response.
This should help dealing with the increase of 1.18 experimental snapshot reports, but at the same time allows keeping other reports which have other future versions only as Awaiting Response. Additionally the specific message might also help users to report experimental snapshot issues at the correct place.

## Approach
The config for the `FutureVersion` module has now a `resolveAsInvalidMessages` `Map<String, String>` for specifying the mapping from specific future version Jira IDs to respective message keys.
Currently the config contains an entry for the 1.18 future version to the `experimental-snapshot` helper message.

## Open question
Is the approach of assuming that future version 1.18 means the user plays the experimental snapshot correct, or could it confuse new users who just happened to select that version?
Should we maybe adjust `experimental-snapshot` or add a slightly modified variant which also tells the user that they can create a new report with the correct version?

## Future work
Remove mappings again from the config once a future version has been released.
However, it does not cause any issues when mappings are not removed immediately after a future version has been released, the module will simply ignore the mappings.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -190,6 +190,9 @@ arisa:
     futureVersion:
       message: provide-affected-versions
       panel: panel-future-version
+      resolveAsInvalidMessages:
+        # 1.18 experimental snapshots
+        "20215": experimental-snapshot
 
     chk:
       projects:
--- a/docs/Modules.md
+++ b/docs/Modules.md
@@ -114,15 +114,20 @@ Resolves an empty report as `Incomplete`.
 | Name  | `FutureVersion`                                                                  |
 | Class | [Link](../src/main/kotlin/io/github/mojira/arisa/modules/FutureVersionModule.kt) |
 
-Removes future versions from the ticket, and resolves it as `Invalid` when the future version is the only version
-the ticket has.
+Removes future versions from the ticket, and resolves it as `Awaiting Response` when the future version is the only version
+the ticket has. If a custom message is specified for a future version by `resolveAsInvalidMessages` in the
+[config](../config/config.yml), the ticket is resolved as `Invalid` instead.
 
 ### Checks
 - Get all the versions of the ticket that are added after last run by non-`staff` users.
 - Any of those versions is a future version.
 - The project of this ticket has a released version.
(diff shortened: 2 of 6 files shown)


Marcono1234:
@Greymagic27, do you think we need a separate message or can we use `experimental-snapshot` for now (see "Open question" section above)?

Greymagic27:
> @Greymagic27, do you think we need a separate message or can we use `experimental-snapshot` for now (see "Open question" section above)?

I think using the experimental snapshot message will be fine since the majority of the users selecting 1.18 will be doing it because they're on the experimental snapshot

Marcono1234:
> However, we could adapt the message to tell them to make a new report if the issue exists on 1.17?

Sounds good. Do you want to create a pull request for this on mojira/helper-messages?

Greymagic27:
> Sounds good. Do you want to create a pull request for this on mojira/helper-messages? Or should I create a pull request and you review it?

I have that one for making the actual helper message, couldn't we add the notice there and then push it?

Marcono1234:
Do you mean https://github.com/mojira/helper-messages/pull/162? That is for `provide-affected-versions` though, I thought you meant we should adjust `experimental-snapshot`.

I have created https://github.com/mojira/helper-messages/pull/163 now, please let me know what you think about it.

Greymagic27:
> Do you mean https://github.com/mojira/helper-messages/pull/162? That is for `provide-affected-versions` though, I thought you meant we should adjust `experimental-snapshot`.
> 
> I have created https://github.com/mojira/helper-messages/pull/163 now, please let me know what you think about it.

Yep, sorry. I was getting confused. I've left a review on 163 for you

SPGoding (approved):

PR #736, opened 2021-08-07
Don't transfer launcher legacy version to duplicated report

Marcono1234:
## Purpose
In most cases launcher legacy version was selected erroneously by user, so it should not be transferred to the duplicated report.
This was originally raised by @dericksonmark.

## Approach
Support specifying a list of Jira version IDs for `TransferVersionsModule` which should not be transferred. The `config.yml` currently only contains the version ID of MCL version "1.6.93 (legacy)".

@dericksonmark proposed adding a `ARISA_IGNORE_VERSION <version>` tag which can be added to a report to prevent Arisa from transferring that version. However, that might be a bit cumbersome especially since a large amount (or nearly all) reports with "1.6.93 (legacy)" have selected it erroneously.

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/docs/Modules.md
+++ b/docs/Modules.md
@@ -420,7 +420,8 @@ Transfers versions from duplicated tickets to their parents.
 
 ### Checks
 - The ticket was linked as a duplicate after last run.
-- The version was released after the oldest version on the parent.
+- The Jira version ID is not listed in `notTransferredVersionIds` specified in the [config](../config/config.yml)
+  (defaults to empty list, i.e. all versions are transferred).
 - The version doesn't exist in the parent yet.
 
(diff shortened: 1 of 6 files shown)


docs/Modules.md
```diff
@@ -420,7 +420,8 @@ Transfers versions from duplicated tickets to their parents.
 
 ### Checks
 - The ticket was linked as a duplicate after last run.
-- The version was released after the oldest version on the parent.
```
Marcono1234, line 423: Removed this because the check had already been removed by #535.

urielsalis (approved):

PR #740, opened 2021-09-25
Switch Future Version to resolving as AR

dericksonmark:
## Purpose
Switch the Future version module to resolving as AR again (basically, reverts #725). I think this should work.
#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


Greymagic27:
This could possibly close #734 unless it's something we still want to keep open

osfanbuff63:
We should probably close #734 because it isn't very relevant if we merge this.

urielsalis (approved):

PR #743, opened 2021-10-05
Add doc for ARISA_MAKE_PRIVATE

### LINKED ISSUE #742: $ARISA_MAKE_PRIVATE is not documented
## The Problem
[Commands.md](https://github.com/mojira/arisa-kt/blob/61c4b4c2bc3ac05d3d855b465478620f3d8dd0dc/docs/Commands.md) does not list info about the command that was added in #669.

SPGoding:
## Purpose

Resolve #742.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -71,6 +71,14 @@ version, the command fails. To skip this check, run with a trailing `force` lite
 
 List all recent activity (up to 50 items) by the given user in a new mod+ comment.
 
+## $ARISA_MAKE_PRIVATE
+| Entry       | Value                       |
+| ----------- | --------------------------- |
+| Syntax      | `$ARISA_MAKE_PRIVATE` |
+| Permissions | Helper+                     |
+
+Marks the ticket as private.
+
 ## $ARISA_PURGE_ATTACHMENT
 | Entry       | Value                                                      |
 | ----------- | ---------------------------------------------------------- |


docs/Commands.md
```diff
@@ -71,6 +71,14 @@ version, the command fails. To skip this check, run with a trailing `force` lite
 
 List all recent activity (up to 50 items) by the given user in a new mod+ comment.
 
+## $ARISA_MAKE_PRIVATE
+| Entry       | Value                       |
+| ----------- | --------------------------- |
+| Syntax      | `$ARISA_MAKE_PRIVATE` |
```
Marcono1234, line 77:
```suggestion
| Syntax      | `$ARISA_MAKE_PRIVATE`       |
```

docs/Commands.md
```diff
@@ -71,6 +71,14 @@ version, the command fails. To skip this check, run with a trailing `force` lite
...
+| Permissions | Helper+                     |
+
+Marks the ticket as private.
```
Marcono1234, line 80: Might be worth explaining why this command exists; what do you think?
```suggestion
Causes Arisa to make the ticket private. This command allows helpers to quickly make a sensitive ticket private
without having to wait for a moderator. Moderators don't need to use this command, they can directly change
the Security Level.

This command only has an effect once; the reporter is able to make the ticket public afterwards again.
```

PR #751, opened 2021-11-25
Make bug reports with attached JFR files private, as they can contain access tokens

urielsalis:
## Purpose
Block JFR files as they contain access tokens
## Approach
Add it to the extension blacklist
#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -107,7 +107,6 @@ arisa:
         - '.rtf'
         - '.pptx'
         - '.sys'
-        - '.jfr'
       comment: attach-new-attachment
 
     duplicateMessage:
@@ -184,6 +183,7 @@ arisa:
         - launcher_accounts.json
         - launcher_msa_credentials.json
         - launcher_profiles.json
+        - .jfr
 
     removeTriagedMeqs:
       resolutions:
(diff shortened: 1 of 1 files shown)


violine1101 (approved):

config/config.yml
```diff
@@ -183,6 +183,7 @@ arisa:
         - launcher_accounts.json
         - launcher_msa_credentials.json
         - launcher_profiles.json
+        - .jfr
```
Marcono1234, line 186: Does this actually work? If I see that correctly `PrivacyModule` checks whether the file names match exactly.

urielsalis: I actually saw a contains

urielsalis: https://github.com/mojira/arisa-kt/blob/master/src/main/kotlin/io/github/mojira/arisa/modules/PrivacyModule.kt#L55

Marcono1234: But that checks if any of the attachment names is contained in the set of `sensitiveFileNames`, isn't it?

We could probably change `sensitiveFileNames` to use regex patterns though, or add a separate `sensitiveFileExtensions`, but that might clutter the class a bit.

PR #754, opened 2021-12-04
Add shadowban command

violine1101:
## Purpose
Add a shadowban command to revert all actions for a specific user, for a duration of 24 hours.

## Approach
Save shadowbans in last run file. I've completely reworked the last run file on the occasion. Have a module that checks for any actions of shadowbanned users on any updated ticket, by looking it up in the last run file essentially. The command just adds the ban to the last run file.

See the commit messages of the individual commits for more information.

There are some shady things I did, please review thoroughly. If you have a better idea on how to solve the communication between command and last run file, feel free to provide suggestions. I felt like threading through the last run file just for this was a bit much.

## Future work
General clean up.

#### Checklist
- [x] Included tests
- [x] Tested in WEB-5505

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@ arisa.yml
 # Files and folders created when running Arisa
 /logs/
 last-run
+lastrun.json
 helper-messages.json
 mc-mappings/
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -155,3 +155,15 @@ Keys and type are case insensitive
 
 Reopens tickets resolved as Awaiting Response
 
+## $ARISA_SHADOWBAN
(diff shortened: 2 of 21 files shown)


urielsalis:
.

urielsalis (approved):

SPGoding (approved):
Looks good to me. As an addicted abuser of global objects I would've added `ShadowBanService` like `HelperMessageService` and have the module and the command communicate through that object lol

violine1101:
One refactor that might be a good idea would be that we just call every module with a `ModuleExecutionContext` or something similar that would contain lastrun, shadowbans, the issue, etc.

PR #757, opened 2021-12-14
Note greedy string arguments in docs

SPGoding:
## Purpose

Make it clear what should be put for some greedy string arguments.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -26,6 +26,8 @@ Keys and type are case insensitive
 | Syntax      | `$ARISA_ADD_VERSION <version>` |
 | Permissions | Helper+                        |
 
+- `<version>`: Put the name of the version on Mojira as-is, including potential spaces. Do not put quotes around it.
+
 ## $ARISA_CLEAR_PROJECT_CACHE
 | Entry       | Value                          |
 | ----------- | ------------------------------ |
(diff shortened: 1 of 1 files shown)


docs/Commands.md
```diff
@@ -116,7 +124,9 @@ Note: the numeric ID of an attachment can be found in the change log or in the l
 | Syntax      | `$ARISA_REMOVE_COMMENTS <username>` |
 | Permissions | Mod+                                |
 
-Finds all comments created by the user on a specific report and restricts them
+- `<username>`: Put the username of the user on Mojira as-is, including potential spaces. Do not encode spaces with the plus sign (`+`). Do not put quotes around it.
+
+Finds all comments created by the user on the current report and restricts them
```
SPGoding, line 129: `on a specific report` is changed to `on the current report` here to make it more _specific_... I hope.

Pokechu22: Since this is already being changed, I think it would make sense to add a trailing period.

Pokechu22:
It might be useful to explicitly note something like "copy the username from the username field, not the URL" (since that's where the pluses came from).

Pokechu22 (approved):

violine1101 (approved):

PR #761, opened 2022-01-06
Format removal message correctly with command prefix

chandler05:
## Purpose
Format removal message correctly with command prefix ($ARISA)

(Leaves the old one there as well just in case someone forgets the $ in front of it and to remove it in tags)

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -296,6 +296,7 @@ arisa:
       removalReason: Comment was not properly staff-restricted.
       removablePrefixes:
         - MEQS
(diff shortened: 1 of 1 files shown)


SPGoding (approved):
(Just FYI there are still things using the `ARISA` prefix as well, like `ARISA_DELETE` or `ARISA_NO_DUPE_MSG`)

chandler05:
Ah yes, I forgot tags and commands have different prefixes from each other.

PR #765, opened 2022-05-08
Update mc-crash-lib to 2.0.7

### LINKED ISSUE #764: Arisa incorrectly closes bug reports if level was modded
## The Bug
Arisa closes bug reports as invalid if the world was opened with a mod, it shouldn't do that.

## Example Ticket
https://bugs.mojang.com/browse/MC-250632

## Expected behavior
Maybe only leave a comment like "Note: World save file has been opened with a mod" but not close the bug report

urielsalis:
## Purpose
Fixes #764

## Approach
Upgrades mc-crash-lib to the fixed version

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-    kotlin("jvm") version "1.4.30"
+    kotlin("jvm") version "1.6.21"
     id("org.jlleitschuh.gradle.ktlint") version "9.2.1"
     application
     id("io.gitlab.arturbosch.detekt") version "1.17.0"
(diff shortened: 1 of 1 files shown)


urielsalis:
Might take around 1 or 2 hours for the repos to update with the latest version, will rerun the check then but it can be reviewed now

Marcono1234 (approved):
Thanks a lot for taking care of this!

Could you please for the sake completeness add the Git tag `2.0.7` to mc-crash-lib?

PR #766, opened 2022-05-12
Revoke unauthorized changes to "Mojang Priority" field

violine1101:
## Purpose
When it's possible to change Mojang Priority despite not being authorized to do so, the bot should revert those changes.

## Approach
Copied `RevokeConfirmationModule`, and changed a few things up.

The mojang priority field appears to be a bit tricky because of the "None" default that needs to be dealt with.
Instead of the name of the field values, I use the IDs which is safer and also means I can revert priority back to "None".

## Future work
Fix up how we interact with JIRA API, this is really janky.

#### Checklist
- [ ] Included tests
- [x] Tested in MCTEST-xxx


violine1101:
Works for newly created bug reports too, just tested it.

chandler05 (approved):

PR #769, opened 2022-05-15
Disable RevokePriorityModule

violine1101:
The RevokePriorityModule needs some more work and throws errors currently, so let's just disable it for now.

#### Checklist
- [ ] Included tests
- [ ] Tested in MCTEST-xxx


osfanbuff63 (approved):
This should work right?

SPGoding (approved):

PR #775, opened 2022-07-02
Detect profilekeys attachments as sensitive

Marcono1234:
## Purpose
The JSON files in the `profilekeys` directory of the game installation directory contain RSA private keys used for authentication at
servers and for signing chat messages. Therefore they should be treated as sensitive.

## Approach
- Update `sensitiveTextRegexes` in config
- Treat MIME types `application/json` (and `application/xml`) as text

#### Checklist
- [x] Included tests
- [ ] Tested in MCTEST-xxx


Greymagic27 (approved):
Looks all fine to me :D

osfanbuff63 (approved):
I would question this since #751, but `private_key` seems like it would be sensitive :P

PR #781, opened 2022-10-03
Improve formatting

Marcono1234:
## Purpose

Follow-up for #779
Fixes some cases of bad line wrapping in the documentation and makes the tests more compact again.


src/test/kotlin/io/github/mojira/arisa/modules/HelpersTest.kt
```diff
@@ -564,33 +450,25 @@ class HelpersTest : StringSpec({
...
-                "1",
-                "1 2",
-                "1 2 3",
-                "2",
-                "2 3",
-                "3"
-            )
-            )
+        LinkType(listOf("1", "2", "3"), "123", true).nameVariants
+            .shouldContainExactly("1", "1 2", "1 2 3", "2", "2 3", "3")
```
Marcono1234, line 471: Changed this to `shouldContainExactly` because that makes sure the collection contains these elements in the specified order. `shouldBe sortedSetOf(...)` apparently would have performed a regular equality check ignoring order. For example `setOf("b", "a") shouldBe sortedSetOf("a", "b")` would pass.

urielsalis (approved):

PR #782, opened 2022-11-19
Rename GitHub workflow files and allow manual triggering of deployment

Marcono1234:
## Purpose
- Renames the GitHub workflow files to express better what they are doing and in which situations they are running (feedback for better names is welcome though)
- Allow manual re-deployment; might be useful in certain situations, see a4156d76b192315686b17888e824e746884a7091
@violine1101, what was actually the context for that re-deployment?

### DIFF
--- a/.github/workflows/pull-request-build.yml
+++ b/.github/workflows/pull-request-build.yml
@@ -1,7 +1,4 @@
-# This workflow will build a Java project with Gradle
-# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
-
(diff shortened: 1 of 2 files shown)


violine1101:
The context for the redeployment was that Arisa no longer saves logs or sends them to Discord. We assumed it was due to a dependency not being added correctly. That turned out not to be the case. - I was looking into that issue briefly but couldn't find the cause.

Actually forgot to file an issue for that - it's now #783.

violine1101 (approved):

osfanbuff63 (approved):

PR #789, opened 2023-04-18
Add Minecraft Legends project

violine1101:
## Purpose
Add the new Minecraft Legends (MCLG) project to Arisa.

## Approach
Update the config -- to my knowledge no updates are needed anywhere else.

Note that MCLG uses a custom platform field, similar to Dungeons. I've not adjusted this yet and as such the platform modules are disabled for now.

## Future work
Hope for the best and that things don't break.

Enable the platform modules for MCLG soon.

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -6,6 +6,7 @@ arisa:
       - MCPE
       - MCL
       - MCD
+      - MCLG
       - BDS
       - REALMS
       - WEB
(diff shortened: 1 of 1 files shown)


violine1101:
I'll merge this immediately because if it's important that Arisa at least acknowledges the existence of the new project (e.g. for AR auto-reopen). I don't think anything should go wrong with the enabled modules. If it does, please let me know.

PR #796, opened 2023-04-20
Draft: Fix platform handling

violine1101:
## Purpose
Fix intermittent error messages when trying to edit bug reports with an old platform that's no longer valid

## Approach
Update the config to include the new platforms. Other necessary updates not implemented yet

## Future work
Maybe refactor platform altogether, it's quite a mess

#### Checklist

- [ ] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -268,12 +268,11 @@ arisa:
       platformWhitelist:
         - Amazon
         - Android
-        - Gear VR
+        - Chromebook
         - iOS
         - Nintendo Switch
-        - PlayStation 4
-        - Windows 10 - PC
-        - Windows 10 Mobile
+        - PlayStation
+        - Windows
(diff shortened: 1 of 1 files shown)


urielsalis (approved):

urielsalis:
Can we do a bulk change of Windows 10 - PC to Windows to reduce impact?

violine1101:
This bulk change (or rather, rename) already happened. That's the issue - it's not mentioned in any of the changelogs, so the KeepPlatform module retrieves "Windows 10 - PC" as the correct platform, but the bug report is set to "Windows", so it tries to change "Windows" to "Windows 10 - PC" which obviously doesn't work. We'll probably need to add a special case to that module just for this reason.

PR #801, opened 2023-06-03
Upgrade Gradle and Kotlin

Marcono1234:
## Purpose
Upgrade Gradle and Kotlin to the current latest versions. Fix warnings and errors for `build.gradle.kts`.

Uses a [Java toolchain](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support) to specify the desired Java version instead of repeating it in multiple places, and relying on the user having a matching JDK installed.


build.gradle.kts
```diff
@@ -41,15 +41,6 @@ val arrowVersion = "0.10.4"
...
 
 dependencies {
-    implementation(kotlin("stdlib-jdk8") as String) {
-        isForce = true
-        isChanging = true
-    }
-    implementation(kotlin("reflect") as String) {
-        isForce = true
-        isChanging = true
-    }
```
Marcono1234, line 51: Have removed this because [according to the documentation](https://kotlinlang.org/docs/gradle-configure-project.html#dependency-on-the-standard-library) a dependency to `stdlib` is added automatically.

Additionally `isForce` was removed by Gradle (and I am not sure if `isChanging = true` was needed in the first place). An replacement seems to be using ["rich versions"](https://docs.gradle.org/current/userguide/rich_versions.html), in particular `strictly`.
The `isForce = true` was originally added by 7bab16902900867b0bd83a652c18c0bbad95e567, maybe it is not needed anymore at the moment and we can just omit it, as done in this pull request.

build.gradle.kts
```diff
@@ -92,43 +83,38 @@ class ClearDependencies : ComponentMetadataRule {
...
 
-tasks {
-    compileKotlin {
-        kotlinOptions.jvmTarget = "11"
-    }
-    compileTestKotlin {
-        kotlinOptions.jvmTarget = "11"
-    }
+kotlin {
+    jvmToolchain(11)
```
urielsalis, line 87: I wonder if we can upgrade to 17

urielsalis (approved):

build.gradle.kts
```diff
@@ -92,43 +83,38 @@ class ClearDependencies : ComponentMetadataRule {
...
-        html.enabled = false // Disabled due to requirement for kotlinx-html which is still in jcenter
-        xml.enabled = false // checkstyle like format mainly for integrations like Jenkins
-        txt.enabled = false // similar to the console output, contains issue signature to manually edit baseline files
-    }
 }
 
 tasks {
     withType<io.gitlab.arturbosch.detekt.Detekt> {
-        // Target version of the generated JVM bytecode. It is used for type resolution.
-        this.jvmTarget = "11"
```
Marcono1234, line 132: Omitted this because [according to the documentation](https://detekt.dev/docs/1.22.0/gettingstarted/gradle/#using-type-resolution) this is inferred from the JVM toolchain.

build.gradle.kts
Marcono1234: Looks like the build still succeeds; but we would probably also have to update the GitHub workflows to use JDK 17 as well, currently it specifies JDK 11. With the JVM toolchain of this PR Gradle should automatically download a matching JDK. But I hope we can avoid this (and therefore improve build speed) by directly using JDK 17 for the GitHub workflows.

Also, not sure if we need to adjust the server where Arisa is currently running to have JDK 17 installed. I am not familiar with the setup there at all.

Should we try switching to Java 17 in a separate PR?

urielsalis: Server would have to be upgradef first indeed

build.gradle.kts
urielsalis: It was added due to a conflict with a dependency originally

urielsalis (approved):

PR #830, opened 2024-09-28
Make MCL use the same security level as the rest of the projects

urielsalis:
DO NOT MERGE
## Purpose
Make MCL use the same security level as the rest of the projects
#### Checklist

- [ ] Included tests
- [ ] Tested in MCTEST-xxx


violine1101:
LGTM; will approve once Arisa is safely shut down and we're ready to merge

violine1101 (approved):

PR #845, opened 2025-03-07
fix: enable reopen awaiting module

jkazimierczak:
## Purpose

Re-registers the AwaitingResponse module, as it was adapted to the cloud API.

[reopen-by-reporter-comment.webm](https://github.com/user-attachments/assets/123dc4bf-e91a-4b8d-97ec-c8dd498ed847)

## Approach

- Replace deprecated `name` attribute usages with `accountId`,
- Remove the logic of checking whether the user account is new (and accompanied tests),
- Port issue transition logic from old client.

#### Checklist

- [x] Included tests
- [ ] Tested in MCTEST-xxx

### DIFF
--- a/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
+++ b/src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
@@ -30,6 +30,8 @@ import java.time.temporal.ChronoField
 import io.github.mojira.arisa.apiclient.JiraClient
 import io.github.mojira.arisa.apiclient.builders.FluentObjectBuilder
 import io.github.mojira.arisa.apiclient.builders.string
+import io.github.mojira.arisa.apiclient.exceptions.JiraClientException
+import io.github.mojira.arisa.apiclient.models.IssueTransition
 import io.github.mojira.arisa.apiclient.models.Visibility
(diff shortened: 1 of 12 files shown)


src/main/kotlin/io/github/mojira/arisa/infrastructure/jira/Functions.kt
```diff
@@ -232,13 +234,30 @@ private fun applyFluentUpdate(issueKey: String, edit: FluentObjectBuilder) = run
     }
 }
 
+private fun getIssueTransition(issueKey: String, transitionName: String): IssueTransition = runBlocking {
+    val allTransitions = jiraClient.getTransitions(issueKey)
+    val transition = allTransitions.transitions?.firstOrNull { it.name == transitionName }
+
+    if (transition == null) {
+        val availableTransitions = allTransitions.transitions?.joinToString(",\n") { "${it.name} (${it.id})" }
+        throw JiraClientException("Transition $transitionName not found. Available transitions:\n$availableTransitions")
```
jkazimierczak, line 243: Should the absence of given transition be just logged with warning/error level instead of throwing?

urielsalis: No transitions being found sound like an immediate issue we should handle, not sure we would look at warnings/error logs as closely

urielsalis (approved):

